PR #39, opened 2025-10-01
Add parameter binding to algorithms bfs_layers, and descendants_at_distance

acarbonetto:
### Summary :memo:
_Write an overview about it._

### Permissions

### DIFF
--- a/example/test_algo_bfs.py
+++ b/example/test_algo_bfs.py
@@ -12,8 +12,6 @@
 # language governing permissions and limitations under the License.
 
 import pytest
-from dotenv import load_dotenv
-load_dotenv()
 
 import networkx as nx
 from nx_neptune import NeptuneGraph, NETWORKX_GRAPH_ID


example/algo_bfs.py
```diff
@@ -10,6 +10,9 @@
 # distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
+from dotenv import load_dotenv
+load_dotenv()
```
andy-k-improving, line 14: Is there any particular reason to have load_dotenv() here rather than line 19, which is after all imports?

andy-k-improving: Also why only this example in particular need `load_dotenv()`, but not all other example.
I think we should let user to specify `env` in their own way rather than mandate the load of `.env`

nx_neptune/algorithms/traversal/bfs.py
```diff
@@ -178,7 +178,11 @@ def descendants_at_distance(
     if concurrency is not None:
         parameters[PARAM_CONCURRENCY] = concurrency
 
-    query_str, para_map = descendants_at_distance_query(source, parameters)
+    source_node = "n"
+
+    query_str, para_map = descendants_at_distance_query(
+        source_node, {f"id({source_node})": source}, parameters
```
andy-k-improving, line 184: I don't see a strong reason to ask for `where filter` while the job of `descendants_at_distance_query()` is to compose the query. 
`{f"id({source_node})": source}` logic can probably move inside of `descendants_at_distance_query()`

nx_neptune/algorithms/traversal/bfs.py
```diff
@@ -228,7 +232,11 @@ def bfs_layers(
     if concurrency is not None:
         parameters[PARAM_CONCURRENCY] = concurrency
 
-    query_str, para_map = bfs_layers_query(sources, parameters)
+    source_node = "n"
```
andy-k-improving, line 235: can we use `_NODE_REF` instead?

example/algo_bfs.py
```diff
@@ -166,12 +176,12 @@
...
 g = nx.path_graph(5)
 logger.info("\n---------scenario: AWS - bfs_layers----------\n")
-result = list(nx.bfs_layers(g, backend=BACKEND, sources=[1, 4]))
+result = list(nx.bfs_layers(g, backend=BACKEND, sources=["1", "4"]))
 logger.info(result)
-assert result == [['4', '1'], ['0', '3', '2']]
+assert result == [['4', '1'], ['2', '3', '0']]
 
 logger.info("\n---------scenario: AWS - bfs_layers - AWS-specific Options----------\n")
-result = list(nx.bfs_layers(g, backend=BACKEND, sources=[1, 4], edge_labels=["RELATES_TO"], traversal_direction="both", concurrency=0))
```
andy-k-improving, line 174: I get a different order of the result, if the order of the result are not deterministic, perhaps we should assert as `set` instead?

Traceback (most recent call last):
  File "/example/algo_bfs.py", line 179, in <module>
    assert result == [['4', '1'], ['2', '3', '0']]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError
INFO - [['4', '1'], ['2', '0', '3']]

nx_neptune/algorithms/traversal/bfs.py
acarbonetto: I'll take a look.

nx_neptune/algorithms/traversal/bfs.py
acarbonetto: sure

PR #48, opened 2025-11-12
Static method - Start / stop instance

andy-k-improving:
### Summary :memo:
_Write an overview about it._

### Permissions

### DIFF
--- a/nx_neptune/instance_management.py
+++ b/nx_neptune/instance_management.py
@@ -311,9 +311,7 @@ def start_na_instance(graph_id: str):
         asyncio.create_task(_wait_until_task_complete(na_client, fut), name=graph_id)
         return asyncio.wrap_future(fut)
     else:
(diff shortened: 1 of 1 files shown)


nx_neptune/instance_management.py
```diff
@@ -274,6 +284,64 @@ def create_na_instance(config: Optional[dict] = None):
...
+        asyncio.Future: A Future that resolves with the graph_id,
+        represent the instance being deleted, or String literal Fail
+        in the case of exception.
+
+    Raises:
+        Exception: If the Neptune Analytics instance creation fails
+    """
+    # Instance deletion
+    na_client = boto3.client("neptune-graph")
+    response = na_client.stop_graph(graphIdentifier=graph_id)
```
acarbonetto: should we validate the state of the graph before doing this?  or should we just assume that the user will check first, and that the stop_graph call will handle the state correctly.  

What happens if the graph is already stopped?  Should we report an error in that case or just return a success with no waiting required?

PR #61, opened 2025-12-04
Make instance_management static functions all async

acarbonetto:
### Summary :memo:

The instance_management.py static functions (mostly) return Futures.  This is a good design, but a more pythonic solution would be to use the `async` flag on the methods and return coroutines.  This allows users to call these methods in parallel using asyncio library.  0

### Test plan:

Units tests are updated to pass. 
Example scripts included to test all scenarios with a live account. 
Session Manager is updated. 
TODO: follow-up PR with notebook examples using SessionManager and NetworkX.  

### Permissions

### DIFF
--- a/nx_neptune/session_manager.py
+++ b/nx_neptune/session_manager.py
@@ -138,6 +138,27 @@ class SessionManager:
                 return graph
         return None
 
+    def get_graph(self, graph_id: str):
+        """Get details for a specific graph by ID.
+
+        Args:
+            graph_id (str): ID of the graph to retrieve
+
+        Returns:
+            dict: Graph details if found
+
+        Raises:
+            Exception: If no graph is found with the given ID
+        """
+        graphs = self.list_graphs()
+
+        for graph in graphs:
+            if graph["id"] == graph_id:
+                return graph
+
+        # Package that as nx object
+        raise Exception(f"No graph instance with id {graph_id} found")
+
     async def get_or_create_graph(self, config: Optional[dict] = None):
         """Get the first available graph or create a new one if none exist.
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -49,7 +49,7 @@ developer = [
     "mypy",
     "sphinx",
     "isort",
-    "pip-licenses"
+    "pip-licenses>=5.5"
 ]
 test = [
     'pytest>=7.2',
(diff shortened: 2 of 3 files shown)


andy-k-improving:
While I understand the rationale for making all session_manager calls asynchronous—since those methods primarily orchestrate graph operations and benefit from coroutine-based composition—I don’t think the same reasoning applies to instance_management.

In instance_management, the coroutine model changes the execution semantics: the method logic is only triggered upon await. This is fundamentally different from the existing Future-based model, where the operation is triggered immediately and await is only used to wait for the result. As a consequence, this change breaks the current use case where users rely on multiple operations being launched concurrently at call time.

In other words, this is not just a syntactic refactor from Future to async; it alters the execution and scheduling behavior in a way that affects concurrency expectations.

```
If the user intends to:
 - Stop an instance (Instance A)
 - Import data from s3 while waiting (Instance B)
 - While stop A is done, run algorithm


Which under the current code (Without async) can do:

future_stop_result = stop_na_instance(instance A)
-- Import on instance B while waiting for A being stop
future_import_result = import_csv_from_s3(Instance B)

await future_stop_result
await future_import_result

-- Only run algo on B when above are done. 
.....some more API call on B

Which total elapsed time will be: (stop_na_instance) + (import_csv_from_s3) - common wait time for both API call


However under the async programming model, since the wait is being encapsulate inside method, and the logic will ONLY be trigger upon await, the below will force to wait in sequential manner, because import_csv_from_s3 will only be triggered after stop_na_instance is done. 

await stop_na_instance(instance A)
await import_csv_from_s3(Instance B)

.....some more API call on B


```

acarbonetto (commented):
Added more exampel tests

example/instance_management/athena_table_import_export.py
```diff
@@ -160,11 +160,13 @@ async def do_export_to_iceberg_table():
 
     iceberg_table_name = 'transactions_updated'
     csv_table_name = 'AwsDataCatalog.bank_fraud_full.transactions_csv_edges'
-    create_iceberg_table_from_table(s3_sql_output, iceberg_table_name, csv_table_name, catalog=catalog, database=database)
+    query_id = await create_iceberg_table_from_table(s3_sql_output, iceberg_table_name, csv_table_name, catalog=catalog, database=database)
+    print(f"Created iceberg table {iceberg_table_name} with query ID: {query_id}")
 
     iceberg_table_name = 'customers_updated'
```
andy-k-improving: let's not to re-use the variables for `iceberg_table_name`, `csv_table_name`, `query_id` and rename it different for second occasion, compare to line:161-163

PR #62, opened 2025-12-09
FEAT: SessionManager.get_graph()

andy-k-improving:
### Summary :memo:
_Write an overview about it._

### Permissions

### DIFF
--- a/nx_neptune/session_manager.py
+++ b/nx_neptune/session_manager.py
@@ -138,6 +138,27 @@ class SessionManager:
                 return graph
         return None
 
+    def get_graph(self, graph_id: str):
+        """Get details for a specific graph by ID.
+
+        Args:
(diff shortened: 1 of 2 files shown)


nx_neptune/session_manager.py
```diff
@@ -138,6 +138,27 @@ def _get_existing_graph(self, filter_status: list[str] | None = None):
...
+
+        Returns:
+            dict: Graph details if found
+
+        Raises:
+            Exception: If no graph is found with the given ID
+        """
+        graphs = self.list_graphs()
+
+        for graph in graphs:
```
acarbonetto: consider one-lining this: 
```
https://stackoverflow.com/questions/2361426/get-the-first-item-from-an-iterable-that-matches-a-condition
```

acarbonetto (approved):

PR #63, opened 2025-12-10
FEAT: Instance management - Resize graph API

andy-k-improving:
### Summary :memo:
* To expose additional call to allow user to size up / down of a graph with ID provided. 

### Permissions

### DIFF
--- a/nx_neptune/instance_management.py
+++ b/nx_neptune/instance_management.py
@@ -57,8 +57,6 @@ logger = logging.getLogger(__name__)
 
 _PROJECT_IDENTIFIER = "nx-neptune"
 
-_ENV_SIZE_LIMIT = "NETWORKX_GRAPH_SIZE_LIMIT"
-
 
 async def create_na_instance(
     config: Optional[dict] = None,


nx_neptune/instance_management.py
```diff
@@ -569,6 +573,73 @@ def stop_na_instance(
...
+
+    Raises:
+        Exception: If the resize operation fails or exceeds size limits
+        ValueError: If the role lacks required permissions or NETWORKX_GRAPH_SIZE_LIMIT is not set
+    """
+    (iam_client, na_client) = _get_or_create_clients(sts_client, iam_client, na_client)
+    # Permission check
+    iam_client.has_update_na_permissions()
+
+    raw_size_limit = os.getenv(_ENV_SIZE_LIMIT)
```
acarbonetto: why is this an environment variable?  can we default this to the maximum ncu size?

nx_neptune/instance_management.py
```diff
@@ -569,6 +573,73 @@ def stop_na_instance(
...
+        )
+        return _get_exception("Exceed size limit")
+
+    logger.info(f"Resizing graph: {graph_id} with size: {prospect_size}")
+    response = na_client.update_graph(
+        graphIdentifier=graph_id, provisionedMemory=prospect_size
+    )
+    status_code = _get_status_code(response)
+    if status_code == 200:
+        return _get_status_check_future(na_client, TaskType.UPDATE, graph_id)
```
acarbonetto: shoudl we make this an await operation?

PR #68, opened 2025-12-30
Cleanup S3 buckets after Session Management call

acarbonetto:
### Summary :memo:

When creating a view from an S3 table, we create an intermediary S3 table to hold the CSV data for import to Neptune.  Once the process is complete, we should ensure that the S3 bucket is deleted.  

A flag is available to turn this feature off. 

Permissions checks are added to all instance_management operations. 

### Test plan:

Through session manager notebooks, and unit tests. 

### Permissions

### DIFF
--- a/nx_neptune/instance_management.py
+++ b/nx_neptune/instance_management.py
@@ -1115,6 +1115,9 @@ async def export_athena_table_to_s3(
     # Check Athena and S3/KMS permissions
     iam_client_wrapper.has_athena_permissions(s3_bucket, key_arn)
 
+    # TODO: In json mode, the annotated embedding field should be handled via
+    #  array_join(transform("test_arr", x -> cast(x as varchar)), ';')
+
     query_execution_ids = []
     for query in sql_queries:
         query_execution_id = _execute_athena_query(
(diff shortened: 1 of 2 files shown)


nx_neptune/instance_management.py
```diff
@@ -1463,9 +1562,79 @@ def _execute_athena_query(
...
+                        response = s3_client.delete_objects(
+                            Bucket=bucket_name,
+                            Delete={"Objects": objects, "Quiet": False},
+                        )
+                        print(
+                            f"Deleted {len(objects)} objects from folder: {bucket_name}/{bucket_path_str}"
+                        )
+        else:
+            # Delete specific key
+            response = s3_client.delete_objects(
```
andy-k-improving: Does it worth an async? Or this delete operation completed relative fast?

nx_neptune/instance_management.py
```diff
@@ -1422,12 +1467,66 @@ async def create_table_schema_from_s3(
...
+
+    # Check the final status
+    final_status = future.current_status
+
+    if final_status != "SUCCEEDED":
+        raise Exception(
+            f"Query {query_execution_id} failed with status: {final_status}"
+        )
+
+    logger.info(f"Successfully completed execution of query [{query_execution_id}]")
```
andy-k-improving: This is sunny case scenario, I wonder does it worth an `info`, lower one level maybe? (Optional)

nx_neptune/instance_management.py
```diff
@@ -1422,12 +1467,66 @@ async def create_table_schema_from_s3(
     return query_execution_id
 
 
+async def drop_athena_table(
+    table: str,
+    s3_bucket: str,
```
andy-k-improving: Why we need the `bucket` while the data deletion is handled by `empty_s3_bucket()`

nx_neptune/instance_management.py
```diff
@@ -1463,9 +1562,79 @@ def _execute_athena_query(
...
+        else:
+            # Delete specific key
+            response = s3_client.delete_objects(
+                Bucket=bucket_name,
+                Delete={
+                    "Objects": [{"Key": bucket_path_str}],
+                    "Quiet": False,
+                },
+            )
+            print(f"Deleted specific key: {bucket_name}/{bucket_path_str}")
```
andy-k-improving: swap out the `print()`

nx_neptune/instance_management.py
```diff
@@ -1422,12 +1467,66 @@ async def create_table_schema_from_s3(
...
 
+async def drop_athena_table(
+    table: str,
+    s3_bucket: str,
+    catalog: str = None,  # type: ignore[assignment]
+    database: str = None,  # type: ignore[assignment]
+    polling_interval=None,
+    max_attempts=None,
+) -> str:
+    """Create external table in Athena from S3 data.
```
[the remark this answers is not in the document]
acarbonetto: Fixed.

nx_neptune/session_manager.py
```diff
@@ -369,6 +376,7 @@ async def import_from_table(
         sql_queries,
         catalog=None,
         database=None,
+        remove_buckets=True,
```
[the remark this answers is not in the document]
acarbonetto: I think it's good practice to cleanup after ourselves.

PR #78, opened 2026-02-05
Fix  Remove parameters from import notebook

acarbonetto:
### Summary :memo:

The import notebook uses athena execution parameters that are irrelevant to the example and should be removed.  

### Test plan:

Integration test (example/instance_management/athena_table_import_export.py) has been updated to include SQL parameters.  

### Permissions

### DIFF
--- a/example/instance_management/athena_table_import_export.py
+++ b/example/instance_management/athena_table_import_export.py
@@ -60,7 +60,7 @@ FROM (
     FROM transactions
     WHERE "nameDest" IS NOT NULL
 )
-WHERE ("type" = ? OR "type" = ? OR "type" = ? OR "type" = ?)
+WHERE "type" IN (?, ?, ?, ?)
 """
 
 BANK_TRANSACTIONS = """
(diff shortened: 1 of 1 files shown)


example/instance_management/athena_table_import_export.py
```diff
@@ -52,14 +52,15 @@
...
     FROM transactions
     WHERE "nameOrig" IS NOT NULL
     UNION ALL
-    SELECT "nameDest" as "~id"
+    SELECT "nameDest" as "~id", "type"
     FROM transactions
     WHERE "nameDest" IS NOT NULL
-);
+)
+WHERE ("type" = ? OR "type" = ? OR "type" = ? OR "type" = ?)
```
andy-k-improving: It's better to do type in (a, b....), if its always around `type`.

acarbonetto: Another problem is that parameters have to be strings, and so I couldn't pass a list of parameters.  ie, I can't do:
`type IN ?`

PR #79, opened 2026-02-09
S3 Vector Connector - Passing embedding as Array[Float]

andy-k-improving:
### Summary :memo:

 - Update license header
 - Update `embedding` data type to array[float]

### Test plan:

Install the connector and confirm embedding is now returning array[float] instead of String.

### Permissions


andy-k-improving:
Merging this feature branch, as there will be another round of review validate this feature end to end.

PR #86, opened 2026-02-11
S3 connector - Implementation of show databases and tables command

andy-k-improving:
### Summary :memo:

Implement the respective method in order to show list of available S3 vector bucket and indexes for external audience (End-user) and internal Athena engine consumption (During query planning runtime).

High level code changes:
* Fetch buket and indexes info from S3 vector API call
* Implement caching for both with default 30s

Caching is being added as both endpoints will also be called by Athena engine internally during the query planning phase, having a short live cache can reduce additional round trip for both API calls to S3 vector service. 

### Permissions


acarbonetto (approved):
Hardcoded cache timeout is fine.  Maybe add it to the constructor so that the user can set it if they like?

PR #88, opened 2026-02-18
S3 Vector connector - Github Action

andy-k-improving:
### Summary :memo:

* Resolve all checkstyle warnings
* Remove the dependency of federation SDK parent POM
* Created a new Github action to trigger artifact build and basic check upon PR submit 

### Permissions

### DIFF
--- a/.github/workflows/athena-s3vector-connector.yml
+++ b/.github/workflows/athena-s3vector-connector.yml
@@ -15,7 +15,7 @@ permissions:
 
 jobs:
   build:
-    runs-on: aws-athena-query-federation_ubuntu-latest_16-core
+    runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


andy-k-improving:
> There are all just checkstyle changes? (except for the workflow)
> 
> We should consider using the athena query federation workflow, since we'd like to bring this into their repository at a future date (or at least keep the option open).
> 
> https://github.com/awslabs/aws-athena-query-federation/blob/master/.github/workflows/maven_push.yml

I copied the portion that applicable to use for now, ex: Build for Java 11 and 17.
But large portion of the job will only make sense when it's running on Federation repo, ex: Centralise code coverage with token......


And yes, all Java changes on this PR are checkstyle fix.

acarbonetto (approved):
lgtm

Issue #103, opened 2026-03-16
Ensure that S3 versioning is enabled


acarbonetto:
**Is your feature request related to a problem? Please describe.**

When using S3 resources, we should validate that the buckets have versioning enabled to ensure that there is no data loss. 

**Describe the solution you'd like**

When checking permissions for S3 usage (write only?), we should also check that hte S3 bucket configuration has versioning enabled. 

**Describe alternatives you've considered**

N/A

**Additional context**
N/A

PR #104, opened 2026-03-16
Test - AWS resources integration tests

andy-k-improving:
### Summary :memo:

Add integration tests for parameterized query safety, graph reset protection, and S3 versioning enforcement

This PR adds integration tests that validate key behaviors against real AWS resources. 

`test_security_injection.py` 
* Inserts nodes and edges with injection payloads against a live Neptune Analytics instance and verifies they are stored as literal data, confirming parameterized queries prevent execution of malicious input. 

`test_security_graph_reset.py` 
* Validates that skip_graph_reset correctly preserves or clears pre-existing graph data. 

`test_security_s3_versioning.py`
* Creates a temporary S3 bucket and confirms check_s3_versioning now raises a ValueError when versioning is not enabled.

### Test plan:

Run integration test locally. 

Close https://github.com/awslabs/nx-neptune/issues/103

### Permissions

### DIFF
--- a/nx_neptune/clients/iam_client.py
+++ b/nx_neptune/clients/iam_client.py
@@ -184,7 +184,9 @@ class IamClient:
             else:
                 raise e
 
-    def check_s3_versioning(self, bucket_arn: str) -> bool:
+    def check_s3_versioning_enabled(
+        self, bucket_arn: str, s3_client: Optional[BaseClient] = None
+    ):
         """Check whether S3 bucket versioning is enabled.
 
         Raises ValueError when versioning is not enabled on buckets used for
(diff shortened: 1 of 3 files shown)


nx_neptune/clients/iam_client.py
```diff
@@ -183,6 +184,36 @@ def check_aws_permission(
             else:
                 raise e
 
+    def check_s3_versioning(self, bucket_arn: str) -> bool:
```
acarbonetto:
```suggestion
    def check_s3_versioning_enabled(self, bucket_arn: str) -> bool:
```

PR #122, opened 2026-03-30
S3 Vector: SAM template permission update

andy-k-improving:
### Summary :memo:

Restrict overly permissive `Resource: '*'` in the S3 Vector connector SAM template to least-privilege scoped ARNs.

**Changes:**
- `athena:GetQueryExecution` — scoped from `*` to `arn:aws:athena:<region>:<account>:workgroup/*`
- S3 Vectors **list access** (`ListIndexes`, `ListVectors`) — scoped from `*` to account/region level for schema and table discovery
- S3 Vectors **data access** (`GetIndex`, `QueryVectors`, `GetVectors`) — scoped to a specific vector bucket and all indexes within it
- Added `S3VectorBucketName` parameter (required) to specify the target vector bucket

All ARNs use CloudFormation pseudo parameters (`AWS::Partition`, `AWS::Region`, `AWS::AccountId`) — no hardcoded values.

**Impact:**
- Lambda can only read vector data from the specified bucket
- Listing operations remain account-wide (required for metadata discovery)
- Cross-account and cross-region access blocked by default
- No functional change for users querying a single vector bucket

### Test plan:
- Deploy connector with `S3VectorBucketName` set to target bucket
- Verify `SHOW SCHEMAS`, `SHOW TABLES`, and `SELECT` queries still work
- Verify queries against other vector buckets are denied

### Permissions

### DIFF
--- a/athena-s3vector-connector/athena-s3vector-connector.yaml
+++ b/athena-s3vector-connector/athena-s3vector-connector.yaml
@@ -52,11 +52,13 @@ Resources:
       FunctionName: !Sub "${AthenaCatalogName}"
       Handler: "com.amazonaws.athena.connectors.s3vector.S3VectorCompositeHandler"
       CodeUri: "./target/athena-s3vector-connector-0.1.0.jar"
-      Description: "A guided example for writing and deploying your own federated Amazon Athena connector for a custom source."
+      Description: "Federated connector to query vector data stored in S3 Vector buckets."
       Runtime: java11
       Timeout: !Ref LambdaTimeout
       MemorySize: !Ref LambdaMemory
       Policies:
+        # It's common and expected for a single Athena connector to serve queries from multiple workgroups.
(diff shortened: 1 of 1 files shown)


athena-s3vector-connector/athena-s3vector-connector.yaml
```diff
@@ -58,20 +61,29 @@ Resources:
             - Action:
                 - athena:GetQueryExecution
               Effect: Allow
-              Resource: '*'
+              Resource: !Sub 'arn:${AWS::Partition}:athena:${AWS::Region}:${AWS::AccountId}:workgroup/*'
```
acarbonetto: Add a comment to say how to restrict this to a specific workgroup if needed.  Something like:
```
It's common and expected for a single Athena connector to serve queries from multiple workgroups.  However, if the lambda is restricted to a specific workgroup update the permission to "arn:aws:athena:<region>:<account-id>:workgroup/<workgroup-name>"
```

PR #181, opened 2026-04-30
Misc: Validate node IDs in openCypher query builder

andy-k-improving:
**Description:**

`_get_nodes_in_list` wraps node IDs in single quotes via string concatenation before interpolating them into `CALL` procedure queries. 
Neptune Analytics's `CALL` procedures don't support parameterized queries (`$param`), so this adds input validation to reject node IDs containing unsafe characters before interpolation.

Node IDs are validated against `^[a-zA-Z0-9_\-:.]+$`, which covers Neptune's system-managed ID formats while rejecting quotes, semicolons, and other injection vectors.

### Changes
- Add regex validation in `_get_nodes_in_list` (`opencypher_builder.py`)
- 3 new unit tests: list of nodes, single node, and injection rejection
- 1 new integration test: closeness centrality with source node list

### Testing
Unit: `326 passed`
Integration: closeness centrality suite passes against live Neptune Analytics instance


tests/clients/test_opencypher_builder.py
```diff
@@ -972,6 +972,31 @@ def test_closeness_centrality_query(self):
...
+        query, params = closeness_centrality_query(source_nodes="A")
+        self.assertIn("'A'", query)
+        self.assertEqual(params, {})
+
+    def test_closeness_centrality_query_rejects_invalid_node_id(self):
+        """Test closeness_centrality_query rejects node IDs with unsafe characters."""
+        from nx_neptune.clients.opencypher_builder import closeness_centrality_query
+
+        with self.assertRaises(ValueError):
+            closeness_centrality_query(source_nodes=["A'; DROP MATCH--"])
```
acarbonetto, line 998: make two separate tests?  we raise an error on the first invalid node

acarbonetto (approved):

tests/clients/test_opencypher_builder.py
andy-k-improving: Done.

PR #238, opened 2026-05-19
Feature: Resource validation module with response utilities

andy-k-improving:
### Summary

Adds upfront resource validation so configuration errors surface immediately — not minutes into a pipeline. Introduces `response_utils.py` as the shared translation layer between raw boto3 responses and typed Python values.

### Architectural context

This PR establishes a layered separation of concerns for the project. `response_utils.py` sits at the infrastructure layer — it owns all boto3 response parsing so that higher-level modules never dig into nested dicts or match error strings directly. `validators.py` sits at the operations layer — it makes API calls, then delegates interpretation to `response_utils` to get definitive answers.

This pattern is designed to scale: as we add more checks or refactor `instance_management`, the same response interpreters and error classifiers are reusable without duplication. New contributors can add a validation check by composing existing building blocks (factory for clients, response_utils for interpretation, CheckResult for output) without understanding boto3 response structures.

### New files

| File | Purpose |
|------|---------|
| `nx_neptune/clients/response_utils.py` | Pure functions: dict in → typed answer out. Response interpreters + error classifiers. |
| `nx_neptune/validators.py` | Resource validation checks using `ClientFactory.default()` and `response_utils` |
| `tests/clients/test_validators.py` | 32 tests, 92% coverage |

### Public API

```python
# Individual checks — return CheckResult
check_bucket_exists("s3://my-bucket/")
check_bucket_region("s3://my-bucket/", "us-west-2")
check_bucket_encryption("s3://my-bucket/")
check_bucket_versioning("s3://my-bucket/")
check_path_empty("s3://my-bucket/output/")
check_athena_database("mydb")
check_athena_table("mydb", "mytable")
check_athena_query("SELECT ...", "mydb", "s3://output/")
check_graph_name_available("my-graph")
check_credentials()

# Orchestrator — returns list[dict]
validate_resources(s3_staging_bucket="s3://...", graph_name="my-graph")
```

### Design

- `response_utils` — stateless module-level functions, no classes. Reusable by `instance_management` and `session_manager` (backlog).
- `CheckResult.ok()` / `.fail()` — concise constructors for pass/fail results
- Error classifiers: `is_not_found`, `is_access_denied`, `is_entity_not_found` — reusable across any check
- Athena query validation uses `wait_until_all_complete` from `task_future` via `asyncio.run()` — no duplicated polling
- All clients from `ClientFactory.default()` — shared cache, single region

### Testing

- 32 validator tests, 92% coverage
- 380 total tests pass, mypy clean

### DIFF
--- a/nx_neptune/clients/response_utils.py
+++ b/nx_neptune/clients/response_utils.py
@@ -11,7 +11,6 @@ from typing import Optional
 
 from botocore.exceptions import ClientError
 
-
 # --- Error classifiers ---
 
 
@@ -34,6 +33,7 @@ def is_entity_not_found(e: ClientError) -> bool:
     """Is this an Athena entity-not-found error?"""
     msg = str(e)
     return "EntityNotFoundException" in msg or "MetadataException" in msg
+
 
 # --- S3 ---
 
--- a/nx_neptune/clients/response_utils.py
+++ b/nx_neptune/clients/response_utils.py
@@ -33,12 +33,6 @@ def is_entity_not_found(e: ClientError) -> bool:
 # --- S3 ---
 
 
-def get_bucket_region(resp: dict) -> str:
-    """Get bucket region from get_bucket_location response."""
-    # S3 API returns None for us-east-1
-    return resp.get("LocationConstraint") or "us-east-1"
-
-
 def is_kms_encrypted(resp: dict) -> bool:
     """Check if get_bucket_encryption response uses KMS."""
     for rule in resp.get("ServerSideEncryptionConfiguration", {}).get("Rules", []):
--- a/nx_neptune/clients/response_utils.py
+++ b/nx_neptune/clients/response_utils.py
@@ -1,0 +1,85 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Response interpreters for boto3 API responses.
+
+Pure functions that translate raw boto3 response dicts into typed answers.
+No API calls — just dict parsing with purpose.
+"""
+
+from typing import Optional
+
+# --- S3 ---
+
+
+def get_bucket_region(resp: dict) -> str:
+    """Get bucket region from get_bucket_location response."""
+    # S3 API returns None for us-east-1
+    return resp.get("LocationConstraint") or "us-east-1"
+
+
+def is_kms_encrypted(resp: dict) -> bool:
+    """Check if get_bucket_encryption response uses KMS."""
+    for rule in resp.get("ServerSideEncryptionConfiguration", {}).get("Rules", []):
+        sse = rule.get("ApplyServerSideEncryptionByDefault", {})
+        if sse.get("SSEAlgorithm") == "aws:kms":
+            return True
+    return False
+
+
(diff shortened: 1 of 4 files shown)


nx_neptune/clients/response_utils.py
```diff
@@ -0,0 +1,106 @@
...
+    msg = str(e)
+    return "EntityNotFoundException" in msg or "MetadataException" in msg
+
+
+# --- S3 ---
+
+
+def get_bucket_region(resp: dict) -> str:
+    """Get bucket region from get_bucket_location response."""
+    # S3 API returns None for us-east-1
```
acarbonetto: ???

nx_neptune/clients/response_utils.py
```diff
@@ -0,0 +1,106 @@
...
+def is_entity_not_found(e: ClientError) -> bool:
+    """Is this an Athena entity-not-found error?"""
+    msg = str(e)
+    return "EntityNotFoundException" in msg or "MetadataException" in msg
+
+
+# --- S3 ---
+
+
+def get_bucket_region(resp: dict) -> str:
```
acarbonetto: not really s3 related...

nx_neptune/validators.py
```diff
@@ -0,0 +1,354 @@
...
+    is_versioning_enabled,
+)
+from nx_neptune.utils.task_future import TaskType, wait_until_all_complete
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass
+class CheckResult:
+    check: str
```
acarbonetto: check_name or check_method?

nx_neptune/validators.py
```diff
@@ -0,0 +1,354 @@
...
+                "s3_bucket_exists", f"Access denied to bucket '{bucket}'"
+            )
+        return CheckResult.fail("s3_bucket_exists", str(e))
+
+
+def check_bucket_region(s3_uri: str, expected_region: str) -> CheckResult:
+    """Check that the bucket is in the expected region."""
+    bucket = _parse_bucket(s3_uri)
+    try:
+        resp = ClientFactory.default().s3().get_bucket_location(Bucket=bucket)
```
acarbonetto: you don't need to wrap subsequent calls after the ClientFactory in a try block.  Better to limit the number of lines for a try block.

acarbonetto (approved):

nx_neptune/clients/response_utils.py
andy-k-improving: This is a documented behaviour under:
https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetBucketLocation.html
Which in the case of us-east-1 will return null.

nx_neptune/validators.py
andy-k-improving: `check_name`.

nx_neptune/clients/response_utils.py
andy-k-improving: Removed, replaced it with inplace check of `s3.head_bucket()`

nx_neptune/validators.py
andy-k-improving: Applied the same all other eligible methods within the module.

PR #241, opened 2026-05-20
Feat: FastAPI proxy server

andy-k-improving:
### Summary

Adds a production-ready FastAPI REST proxy (`proxy/`) that will serve as the HTTP interface for nx-neptune. This layer owns transport, configuration, and operational concerns — all business logic is delegated to the nx-neptune library.

This PR ships the foundational skeleton only. Business endpoints (setup, query, validation, graph management) will be added incrementally in follow-up PRs.

### What's included

```
proxy/
├── Dockerfile              # single-stage, non-root user, wget healthcheck
├── Makefile                # dev, install, test, build, run, stop, clean (finch by default)
├── pyproject.toml          # pinned deps, aligned with nx-neptune style
├── src/nx_neptune_proxy/
│   ├── __init__.py
│   ├── app.py              # FastAPI app, /health, error handler, logging, CORS
│   └── config.py           # Settings from env vars, validated at startup
├── tests/
│   └── test_app.py         # 5 tests
└── ui/
    └── index.html          # minimal page that calls /health
```

### Production concerns addressed

- **Structured logging** — JSON format, request ID correlation on every request
- **Global error handler** — consistent JSON error shape, no stack traces in responses
- **CORS** — locked down by default, configurable via `CORS_ALLOWED_ORIGINS` env var
- **Configuration** — `Settings.from_env()` validates at startup (LOG_LEVEL, CORS_ALLOWED_ORIGINS, PORT)
- **Container** — non-root user, HEALTHCHECK via wget, single stage
- **Dependency pinning** — `fastapi==0.115.12`, `uvicorn==0.34.2`
- **CI** — `proxy_tests` job added to `.github/workflows/main.yml`

### How to run

```bash
cd proxy
make install    # editable install with test deps
make dev        # http://localhost:8080
make test       # run tests
make build      # container image (uses finch, override with RUNTIME=docker)
make run        # run container
```

image

### Testing

- 5 proxy tests pass (health, request-id, error handling, docs)
- 348 main nx-neptune tests unaffected

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -133,6 +133,8 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
       - name: Install proxy with test deps
-        run: pip install -e "./proxy[test]"
+        run: pip install -e ".[test]"
+        working-directory: proxy
       - name: Run proxy tests
-        run: pytest proxy/tests/ -v
+        run: pytest tests/ -v
+        working-directory: proxy
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -119,3 +119,20 @@ jobs:
               issue_number: prNumber,
               body: `📊 **Coverage Report**: Download the HTML report [here](${artifactUrl}).`
             });
+
+  proxy_tests:
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: [3.11]
+        os: [ubuntu-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install proxy with test deps
+        run: pip install -e "./proxy[test]"
+      - name: Run proxy tests
+        run: pytest proxy/tests/ -v
--- a/proxy/src/nx_neptune_proxy/config.py
+++ b/proxy/src/nx_neptune_proxy/config.py
@@ -1,0 +1,28 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+import os
+from dataclasses import dataclass
+
+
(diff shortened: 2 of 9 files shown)


.github/workflows/main.yml
```diff
@@ -119,3 +119,20 @@ jobs:
...
+        python-version: [3.11]
+        os: [ubuntu-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install proxy with test deps
+        run: pip install -e "./proxy[test]"
```
acarbonetto: better to change directory to ./proxy and `pip install .[test]`

proxy/src/nx_neptune_proxy/config.py
```diff
@@ -0,0 +1,28 @@
...
+    allowed_origins: list[str] = None  # type: ignore[assignment]
+    port: int = 8080
+
+    def __post_init__(self) -> None:
+        if self.allowed_origins is None:
+            object.__setattr__(self, "allowed_origins", [])
+
+    @classmethod
+    def from_env(cls) -> "Settings":
+        origins_raw = os.environ.get("CORS_ALLOWED_ORIGINS", "")
```
acarbonetto, line 22: this won't work without the environment variable being set.  would it be good to fail now if the environment variable is not set?

acarbonetto (approved):
not opposed to this PR, but there's a chance we will have to change the ui/index.html substantially.

proxy/src/nx_neptune_proxy/config.py
andy-k-improving: Empty is intentional — when the UI is served from the proxy itself (same-origin), CORS headers aren't needed. Failing on an unset `CORS_ALLOWED_ORIGINS` would break standalone usage where the user doesn't need cross-origin access.
When Graph Explorer or another frontend runs on a different origin, they set the env var and it works. If they forget, the browser shows a clear CORS error — it won't fail silently.

.github/workflows/main.yml
andy-k-improving: Done.

PR #242, opened 2026-05-20
Feat: ClientFactory adoption - pt2

andy-k-improving:
### Summary

Completes the `ClientFactory` adoption across all core modules. Every AWS client in the main code path now goes through `ClientFactory.default()`, ensuring consistent region, session, and caching behavior.

### Changes

| File | What changed |
|------|-------------|
| `nx_neptune/clients/na_client.py` | Uses `ClientFactory.default().neptune()` when no custom timeout; only creates a separate factory for custom `timeout_seconds` |
| `nx_neptune/clients/iam_client.py` | Replaced `boto3.client("s3")` fallback with `ClientFactory.default().s3()`, removed `boto3` import |
| `nx_neptune/na_graph.py` | Replaced `boto3.client(SERVICE_STS)` and `boto3.client(SERVICE_IAM)` with `ClientFactory.default()`, removed `boto3` import and unused `SERVICE_*` constants |

### Why

Before this change, `na_client`, `iam_client`, and `na_graph` created their own boto3 clients independently. This meant:
- No shared caching — duplicate clients for the same service
- No region guarantee — each client could resolve to a different default region
- Inconsistent testing — had to mock boto3 in multiple places

Now there's one door in for AWS access. Mock `ClientFactory.default()` once and the entire library uses it.

### Not in scope

`utils/utils.py` still has 3 direct `boto3.client()` calls for experimental services (`bedrock-runtime`, `s3vectors`) not yet in `ClientFactory`. These are utility functions, not core path.

### Testing

- 348 tests pass, mypy clean

### DIFF
--- a/nx_neptune/clients/client_factory.py
+++ b/nx_neptune/clients/client_factory.py
@@ -34,8 +34,8 @@ class ClientFactory:
     Ensures consistent configuration (user agent, timeouts, region) across
     all AWS service clients used by the library.
 
-    Use ``ClientFactory.default()`` for a shared singleton, or create an
-    instance for custom configuration.
+    When called with no arguments, returns a shared singleton instance.
+    When called with custom configuration, creates a new instance.
 
     Args:
         region: AWS region name. If None, uses boto3 default.
@@ -45,21 +45,29 @@ class ClientFactory:
 
     _default: Optional["ClientFactory"] = None
 
-    @classmethod
-    def default(cls) -> "ClientFactory":
-        """Return the shared default factory (lazy-initialized)."""
-        if cls._default is None:
-            cls._default = cls()
-        return cls._default
+    def __new__(
+        cls,
+        region: Optional[str] = None,
+        timeout_seconds: Optional[int] = None,
+    ):
+        if region is None and timeout_seconds is None:
+            if cls._default is None:
+                cls._default = super().__new__(cls)
+                cls._default._initialized = False
+            return cls._default
+        return super().__new__(cls)
 
     def __init__(
         self,
         region: Optional[str] = None,
         timeout_seconds: Optional[int] = None,
     ):
+        if getattr(self, "_initialized", False):
+            return
         self._region = region
         self._timeout_seconds = timeout_seconds
         self._clients: dict[str, BaseClient] = {}
+        self._initialized = True
 
     def _base_kwargs(self) -> dict[str, Any]:
         kwargs: dict[str, Any] = {}
--- a/nx_neptune/clients/client_factory.py
+++ b/nx_neptune/clients/client_factory.py
@@ -53,7 +53,6 @@ class ClientFactory:
         if region is None and timeout_seconds is None:
             if cls._default is None:
                 cls._default = super().__new__(cls)
-                cls._default._initialized = False
             return cls._default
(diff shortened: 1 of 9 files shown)


nx_neptune/clients/na_client.py
```diff
@@ -57,8 +57,10 @@ def __init__(
         self.graph_id = graph_id
         if client:
             self.client = client
-        else:
+        elif timeout_seconds:
             self.client = ClientFactory(timeout_seconds=timeout_seconds).neptune()
+        else:
+            self.client = ClientFactory.default().neptune()
```
acarbonetto: I don't like this pattern to create the ClientFactory.  The create or get should be abstracted away.  Can we include the get pattern if no parameters are passed in?  
So this: 
```
timeout_seconds=None
self.client = ClientFactory(timeout_seconds=timeout_seconds).neptune()
```
is the same as: 
```
self.client = ClientFactory().neptune()
```
which is the same as (but we don't need and can remove): 
```
ClientFactory.default()
```

andy-k-improving: Updated to remove .default syntax.

nx_neptune/clients/na_client.py
```diff
@@ -57,8 +57,10 @@ def __init__(
         self.graph_id = graph_id
         if client:
             self.client = client
-        else:
+        elif timeout_seconds:
             self.client = ClientFactory(timeout_seconds=timeout_seconds).neptune()
+        else:
+            self.client = ClientFactory().neptune()
```
acarbonetto: can we remove this now?  Use: 
```
self.client = ClientFactory(timeout_seconds=timeout_seconds).neptune()
```
even if timeout_seconds is not defined?

nx_neptune/clients/client_factory.py
```diff
@@ -45,21 +45,29 @@ class ClientFactory:
...
         self,
         region: Optional[str] = None,
         timeout_seconds: Optional[int] = None,
     ):
+        if getattr(self, "_initialized", False):
+            return
         self._region = region
         self._timeout_seconds = timeout_seconds
         self._clients: dict[str, BaseClient] = {}
+        self._initialized = True
```
acarbonetto: what is this used for?   I think we can just check if _default is None, right?

nx_neptune/clients/client_factory.py
```diff
@@ -34,8 +34,8 @@ class ClientFactory:
     Ensures consistent configuration (user agent, timeouts, region) across
     all AWS service clients used by the library.
 
-    Use ``ClientFactory.default()`` for a shared singleton, or create an
-    instance for custom configuration.
+    When called with no arguments, returns a shared singleton instance.
```
acarbonetto:
```suggestion
    When called with no arguments, returns the shared singleton default instance.
```

nx_neptune/clients/na_client.py
andy-k-improving: Done.

nx_neptune/clients/client_factory.py
andy-k-improving: Remove, replaced with _client check.

acarbonetto (approved):

PR #294, opened 2026-06-26
Feat: Add workspace concept for grouping projections

andy-k-improving:
## Summary

Adds a Workspace concept as a logical grouping for projections. Users create workspaces in the sidebar, then create projections (graphs) under them via "+ Add Graph".

### Backend (Python/FastAPI)
- New `workspace_store.py` — Workspace model (id, name, status) with in-memory store
- New `workspace.py` router — CRUD at `/api/v0/workspace`
- New `workspace_deletion.py` — async cascade: delete Neptune graphs → remove projections → remove workspace
- Updated `projection_store.py` — added `workspace_id` field to Projection
- Updated `projection.py` — added `DELETE` endpoint (async graph deletion + record removal)
- Updated `schemas.py` — added `workspace_id` to create/response models
- Updated `app.py` — registered workspace router

### Frontend (React/TypeScript)
- **Sidebar**: collapsible workspace tree showing projections with status dots, "+ Add Graph" link, delete-on-hover, auto-refresh via custom events
- **Import page**: workspace dropdown with inline create, copy-config-from dropdown, reacts to URL params (`?workspace=`, `?session=`)
- **Sessions page**: filters by workspace, shows graph status badge in detail panel, delete button in panel
- **New Workspaces page**: simple name input + create button
- Removed Query Builder page and its nav entry
- Moved Graphs nav link above Workspaces section

Screenshot 2026-06-26 at 11 59 29 AM
Screenshot 2026-06-26 at 12 27 24 PM
Screenshot 2026-06-26 at 12 27 34 PM

## Behavior
- Workspace deletion is async (202) — background task polls Neptune until graphs are gone
- Sidebar refreshes on create/delete via `workspaces-changed` custom event
- Clicking workspace name opens filtered sessions view
- "+ Add Graph" always opens a blank import form with workspace pre-selected

## What was tested
- `npx tsc --noEmit` passes
- `npm run build` succeeds
- Python `from nx_neptune_proxy.app import app` imports cleanly
- Manual verification of endpoints via curl


acarbonetto:
A couple of first impressions: 
* I'm guessing you prefer "Workspace" over "Project", right?  We will probably need to defend the choice.  
* Also, we should have a landing page when there are zero workspaces (so the right panel isn't blank, but instead says something like "Click "+" to create a workspace" (even better if the right panel has a cloned button).  
* When I create a new workspace, it would be nice to have the workspace expanded by default (instead of having to expand it before the "Add graph option is available").  
* If I'm on the import page of a workspace (or any page within a workspace), and I delete the workspace - where does the page go?  
* (out of scope?) if I delete a workspace, I should probably get one of those "are you sure?" pop-ups... especially if the workspace has graphs.  The pop up should may have a checkbox (default: checked) to delete all the graphs in the workspace.

andy-k-improving:
> A couple of first impressions:
> 
> * I'm guessing you prefer "Workspace" over "Project", right?  We will probably need to defend the choice.
> * Also, we should have a landing page when there are zero workspaces (so the right panel isn't blank, but instead says something like "Click "+" to create a workspace" (even better if the right panel has a cloned button).
> * When I create a new workspace, it would be nice to have the workspace expanded by default (instead of having to expand it before the "Add graph option is available").
> * If I'm on the import page of a workspace (or any page within a workspace), and I delete the workspace - where does the page go?
> * (out of scope?) if I delete a workspace, I should probably get one of those "are you sure?" pop-ups... especially if the workspace has graphs.  The pop up should may have a checkbox (default: checked) to delete all the graphs in the workspace.



**Project** 

Renamed from workspace to project across the codebase, both front and backend. 

**Landing page when zero projects exist**

The `/` route now checks project count:
- Zero projects → redirects to the create project page (has a form + button)
- Has projects → redirects to the first project's sessions page


**Auto-expand new project in sidebar**

The sidebar now auto-expands whichever project is referenced in the current URL (`?project=` or `?session=`). So after creating a project and landing on its sessions page, it's already expanded with "+ Add Graph" visible.

**Delete project while on its page**

After deleting a project from the sidebar, if the current page references that project, it navigates to `/` (which triggers the smart redirect — next active project, or create page if none remain). Projects with `status: "deleting"` are filtered out of the redirect logic.

**Project deletion**

Regarding the option to delete project without graph, I prefer to skip it for now, as orphaned graphs (no project association) add confusion without clear value. Cascade delete is always the right behavior here.

proxy/src/nx_neptune_proxy/routers/project.py
```diff
@@ -0,0 +1,68 @@
...
+        raise HTTPException(status_code=404, detail="Project not found")
+    return asdict(p)
+
+
+@router.delete("/{project_id}", status_code=202)
+def delete_project_endpoint(project_id: str, background_tasks: BackgroundTasks):
+    p = store.get(project_id)
+    if p is None:
+        raise HTTPException(status_code=404, detail="Project not found")
+    if p.status == "deleting":
```
acarbonetto, line 64: The delete_project_endpoint mutates the Project dataclass directly instead of going through store.update(). This bypasses any future store-level logic and is inconsistent with how the projection deletion (which uses store.update(...)) works. More importantly, if the background task fails, the status is stuck at "deleting" with no recovery path.

Suggestion: Use store.update(project_id, status="deleting") and consider adding error recovery in the background task (reset status on failure).

proxy/src/nx_neptune_proxy/routers/project.py
```diff
@@ -0,0 +1,68 @@
...
+        raise HTTPException(status_code=409, detail="Already deleting")
+    p.status = "deleting"
+    background_tasks.add_task(asyncio.run, delete_project(project_id))
```
acarbonetto, line 67: asyncio.run() creates a new event loop. If FastAPI is already running an asyncio loop (which it does under uvicorn), calling asyncio.run() from a background thread can work but is fragile — it creates a second loop. The idiomatic approach with FastAPI background tasks for async functions
  
is:
background_tasks.add_task(delete_project, project_id)

FastAPI's BackgroundTasks natively supports coroutines — it will await them directly without needing asyncio.run.

proxy/src/nx_neptune_proxy/routers/projection.py
```diff
@@ -142,3 +142,36 @@ def execute_projection(projection_id: str, background_tasks: BackgroundTasks):
...
+            # Poll until gone
+            for _ in range(60):
+                await asyncio.sleep(10)
+                try:
+                    client.get_graph(graphIdentifier=p.graph_id)
+                except ClientError:
+                    break
+        store.delete(projection_id)
+
+    background_tasks.add_task(asyncio.run, _delete())
```
acarbonetto, line 176: same asyncio issue

proxy/src/nx_neptune_proxy/routers/projection.py
```diff
@@ -142,3 +142,36 @@ def execute_projection(projection_id: str, background_tasks: BackgroundTasks):
...
+
+@router.delete("/{projection_id}", summary="Delete projection and its graph", status_code=202)
+def delete_projection(projection_id: str, background_tasks: BackgroundTasks):
+    """Set status to deleting, delete graph in background, then remove record."""
+    p = _get_projection_or_404(projection_id)
+    if p.status == "deleting":
+        raise HTTPException(status_code=409, detail="Already deleting")
+    store.update(projection_id, status="deleting", step="graph_delete", step_label="Deleting graph")
+
+    async def _delete():
```
acarbonetto, line 155: The inline _delete() coroutine has no logging and only sets status="failed" on the initial delete_graph call. If the polling loop times out (60 iterations × 10s = 10 min), the projection record is deleted anyway via store.delete(projection_id) even though the graph may still exist. This could leave orphaned Neptune graphs.

Suggestion: Add timeout handling that sets the projection to failed with an error message instead of silently deleting the record.

proxy/src/nx_neptune_proxy/routers/project.py
```diff
@@ -0,0 +1,68 @@
...
+class ProjectResponse(BaseModel):
+    id: str
+    name: str
+    status: str
+    created_at: datetime
+
+
+@router.post("", response_model=ProjectResponse, status_code=201)
+def create_project(body: ProjectCreate):
+    return asdict(store.create(name=body.name))
```
acarbonetto, line 35: ProjectCreate accepts any string including empty/whitespace-only names. The frontend guards with name.trim() but the API endpoint doesn't validate.

  Suggestion: Add name: str = Field(min_length=1, max_length=255) or a validator.

proxy/src/nx_neptune_proxy/routers/project.py
```diff
@@ -0,0 +1,68 @@
...
+from fastapi import APIRouter, BackgroundTasks, HTTPException
+from pydantic import BaseModel
+
+from nx_neptune_proxy.services.project_deletion import delete_project
+from nx_neptune_proxy.services.project_store import store
+
+router = APIRouter(prefix="/api/v0/project", tags=["project"])
+
+
+class ProjectCreate(BaseModel):
```
acarbonetto, line 18: consider adding details on what these structs are modelling

acarbonetto (approved):

proxy/src/nx_neptune_proxy/routers/project.py
andy-k-improving: Fixed — now uses store.update(project_id, status="deleting"). Also added startup recovery:
  on boot, any projects stuck in "deleting" get re-queued for deletion.

proxy/src/nx_neptune_proxy/routers/project.py
andy-k-improving: Fixed — changed to background_tasks.add_task(delete_project, project_id). FastAPI natively awaits coroutines in background tasks.

proxy/src/nx_neptune_proxy/routers/projection.py
andy-k-improving: Fixed — both _delete() and run_pipeline() now passed directly to background_tasks.add_task() without the asyncio.run wrapper.

proxy/src/nx_neptune_proxy/routers/project.py
andy-k-improving: Fixed — added name: str = Field(min_length=1, max_length=255) to both ProjectCreate an   ProjectUpdate. Frontend already guards this but now the API rejects it too.

proxy/src/nx_neptune_proxy/routers/projection.py
andy-k-improving: Added logging for time out.

andy-k-improving:
Merged as part of https://github.com/awslabs/nx-neptune/pull/299

PR #311, opened 2026-07-03
Feat: Add graph instance management (start/stop/delete) to Graph Studio

andy-k-improving:
## Summary

Adds the ability to manage Neptune Analytics graph lifecycle directly from the Graph Studio UI — start, stop, and delete graphs with proper state validation and async execution.

## Changes

**Backend — state machine pattern**
- `services/graph_state_machine.py` — Generic transition engine with validation, async polling, in-flight error tracking
- `services/graph_ops.py` — Concrete operations (stop, start, delete) registered as transitions. Adding new actions only requires changes here.
- `routers/graph.py` — REST endpoints: `GET /{graph_id}/actions`, `POST /{graph_id}/{action}`, `GET/DELETE /{graph_id}/inflight`
- `app.py` — Register graph router

**Frontend**
- `api/index.ts` — Added `graphActions` API namespace
- `Graphs.tsx` — Action buttons (start/stop/delete) driven by backend state, disabled when invalid, auto-poll during transient states (30s), error alerts with dismiss

**Infra**
- `docker-compose.yml` — Use `GRAPH_EXPLORER_IMAGE` variable for flexible local testing
- `Makefile` — Added `make up-ge` target for running only Graph Explorer

## Valid transitions

| Current State | Available Actions |
|--------------|-------------------|
| AVAILABLE | stop, delete |
| STOPPED | start, delete |
| Other | none (buttons disabled) |

Screenshot 2026-07-03 at 2 06 06 PM
Screenshot 2026-07-03 at 2 09 12 PM
Screenshot 2026-07-03 at 2 12 27 PM

## Testing

- All 34 existing tests pass
- Python import verification passes
- UI builds successfully

### DIFF
--- a/proxy/src/nx_neptune_proxy/routers/graph.py
+++ b/proxy/src/nx_neptune_proxy/routers/graph.py
@@ -1,0 +1,74 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Router for graph instance actions (start, stop, delete)."""
+
+from __future__ import annotations
+
+from fastapi import APIRouter, BackgroundTasks, HTTPException
+
+from nx_neptune.clients.client_factory import ClientFactory
+from nx_neptune_proxy.services.graph_state_machine import (
+    InvalidTransitionError,
+    available_actions,
+    execute_transition,
+    get_inflight,
+    clear_inflight,
+)
+import nx_neptune_proxy.services.graph_ops  # noqa: F401 — registers transitions
+
+router = APIRouter(prefix="/api/v0/graphs", tags=["graphs"])
+
+
+@router.get("/{graph_id}/actions", summary="List available actions for a graph")
+def get_available_actions(graph_id: str):
+    """Return valid actions based on the graph's current Neptune status."""
+    client = ClientFactory().neptune()
+    resp = client.get_graph(graphIdentifier=graph_id)
+    status = resp.get("status", "")
+    actions = available_actions(status)
+    inflight = get_inflight(graph_id)
+    return {
+        "graph_id": graph_id,
+        "status": status,
+        "actions": actions,
+        "inflight": inflight,
+    }
+
+
+@router.post("/{graph_id}/{action}", summary="Execute a graph action", status_code=202)
+def perform_action(graph_id: str, action: str, background_tasks: BackgroundTasks):
+    """Initiate a state transition (stop, start, delete) as a background task."""
+    client = ClientFactory().neptune()
+    resp = client.get_graph(graphIdentifier=graph_id)
+    current_status = resp.get("status", "")
+
+    try:
+        # Validate early (execute_transition also validates, but fail fast for the user)
+        valid = available_actions(current_status)
+        if action not in valid:
+            raise InvalidTransitionError(
+                f"Cannot {action} graph in {current_status} state. "
+                f"Valid actions: {valid or 'none'}"
+            )
+    except InvalidTransitionError as e:
+        raise HTTPException(status_code=409, detail=str(e))
+
+    background_tasks.add_task(execute_transition, graph_id, action, current_status)
+    return {"graph_id": graph_id, "action": action, "status": "accepted"}
+
+
+@router.get("/{graph_id}/inflight", summary="Check in-flight operation status")
+def get_inflight_status(graph_id: str):
+    """Return current in-flight operation info, including any error."""
+    inflight = get_inflight(graph_id)
+    if not inflight:
+        return {"graph_id": graph_id, "inflight": None}
+    return {"graph_id": graph_id, "inflight": inflight}
+
+
+@router.delete("/{graph_id}/inflight", summary="Dismiss inflight error")
+def dismiss_inflight(graph_id: str):
+    """Clear the in-flight error for a graph (user dismissed the alert)."""
+    clear_inflight(graph_id)
+    return {"graph_id": graph_id, "cleared": True}
--- a/proxy/src/nx_neptune_proxy/routers/graph.py
+++ b/proxy/src/nx_neptune_proxy/routers/graph.py
@@ -5,6 +5,8 @@
 
 from __future__ import annotations
 
+import logging
+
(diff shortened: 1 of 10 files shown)


proxy/src/nx_neptune_proxy/routers/graph.py
```diff
@@ -0,0 +1,74 @@
...
+import nx_neptune_proxy.services.graph_ops  # noqa: F401 — registers transitions
+
+router = APIRouter(prefix="/api/v0/graphs", tags=["graphs"])
+
+
+@router.get("/{graph_id}/actions", summary="List available actions for a graph")
+def get_available_actions(graph_id: str):
+    """Return valid actions based on the graph's current Neptune status."""
+    client = ClientFactory().neptune()
+    resp = client.get_graph(graphIdentifier=graph_id)
```
acarbonetto, line 31: 404 if the graph doesn't exist?

proxy/src/nx_neptune_proxy/routers/graph.py
```diff
@@ -0,0 +1,74 @@
...
+    status = resp.get("status", "")
```
acarbonetto: what does it mean to have the empty state? maybe empty actions and report a warning?

proxy/src/nx_neptune_proxy/routers/graph.py
```diff
@@ -0,0 +1,74 @@
...
+    execute_transition,
+    get_inflight,
+    clear_inflight,
+)
+import nx_neptune_proxy.services.graph_ops  # noqa: F401 — registers transitions
+
+router = APIRouter(prefix="/api/v0/graphs", tags=["graphs"])
+
+
+@router.get("/{graph_id}/actions", summary="List available actions for a graph")
```
acarbonetto, line 27: we should make sure to add unit tests for all public apis

proxy/src/nx_neptune_proxy/routers/graph.py
```diff
@@ -0,0 +1,74 @@
...
+        "inflight": inflight,
+    }
+
+
+@router.post("/{graph_id}/{action}", summary="Execute a graph action", status_code=202)
+def perform_action(graph_id: str, action: str, background_tasks: BackgroundTasks):
+    """Initiate a state transition (stop, start, delete) as a background task."""
+    client = ClientFactory().neptune()
+    resp = client.get_graph(graphIdentifier=graph_id)
+    current_status = resp.get("status", "")
```
acarbonetto: if "" error?

proxy/src/nx_neptune_proxy/routers/graph.py
```diff
@@ -0,0 +1,74 @@
...
+        "actions": actions,
+        "inflight": inflight,
+    }
+
+
+@router.post("/{graph_id}/{action}", summary="Execute a graph action", status_code=202)
+def perform_action(graph_id: str, action: str, background_tasks: BackgroundTasks):
+    """Initiate a state transition (stop, start, delete) as a background task."""
+    client = ClientFactory().neptune()
+    resp = client.get_graph(graphIdentifier=graph_id)
```
acarbonetto, line 50: 404

proxy/src/nx_neptune_proxy/routers/graph.py
```diff
@@ -0,0 +1,74 @@
...
+        raise HTTPException(status_code=409, detail=str(e))
+
+    background_tasks.add_task(execute_transition, graph_id, action, current_status)
+    return {"graph_id": graph_id, "action": action, "status": "accepted"}
+
+
+@router.get("/{graph_id}/inflight", summary="Check in-flight operation status")
+def get_inflight_status(graph_id: str):
+    """Return current in-flight operation info, including any error."""
+    inflight = get_inflight(graph_id)
```
acarbonetto, line 74: consider 404

proxy/src/nx_neptune_proxy/routers/graph.py
```diff
@@ -0,0 +1,74 @@
...
+    if not inflight:
+        return {"graph_id": graph_id, "inflight": None}
+    return {"graph_id": graph_id, "inflight": inflight}
+
+
+@router.delete("/{graph_id}/inflight", summary="Dismiss inflight error")
+def dismiss_inflight(graph_id: str):
+    """Clear the in-flight error for a graph (user dismissed the alert)."""
+    clear_inflight(graph_id)
```
acarbonetto, line 83: consider 404

acarbonetto (approved):

proxy/src/nx_neptune_proxy/routers/graph.py
andy-k-improving: The current `aws_exception_handler` already catching this, in the case of graph doesn't exist, it will throw a ClientError with `ResourceNotFoundException` which will be picked up by `aws_exception_handler` and return 404 instead.

proxy/src/nx_neptune_proxy/routers/graph.py
andy-k-improving: Updated to have the warning and return 502 instead with similar error message.

proxy/src/nx_neptune_proxy/routers/graph.py
andy-k-improving: Tests are being added for new routes.

proxy/src/nx_neptune_proxy/routers/graph.py
andy-k-improving: It's being covered by default exception handler: `aws_exception_handler`

proxy/src/nx_neptune_proxy/routers/graph.py
andy-k-improving: Intentionally a no-op — `DELETE` should be idempotent. If there's nothing to dismiss, returning 200 with `cleared: true` is safe.

proxy/src/nx_neptune_proxy/routers/graph.py
andy-k-improving: `inflight` isn't a resource — it's a property of the graph. Returning `null` means "nothing in progress." A 404 would imply the graph itself doesn't exist.

PR #314, opened 2026-07-06
CI: Ignore JUnit vintage engine in Dependabot config

andy-k-improving:
**Description:**

Add `org.junit.vintage:*` to the Dependabot ignore list for both Maven connectors.
The `org.junit.jupiter:*` ignore was already in place but didn't cover the vintage engine, which allowed PRs like #306 (JUnit 5→6 major bump) to still be created.


andy-k-improving:
Merging this as the build is green and the updated on dependabot can only be tested on main.

PR #330, opened 2026-07-16
Fix: Scope "Copy config from" dropdown to current project

andy-k-improving:
**Description:**

The "Copy config from" dropdown on the import page now only shows projections belonging to the current project. If no project is selected, all projections are shown.

Client-side filter on the existing `projectionsList` state — no backend changes needed.

### DIFF
--- a/proxy/ui-src/src/pages/Import.tsx
+++ b/proxy/ui-src/src/pages/Import.tsx
@@ -253,7 +253,7 @@ export function Import() {
                 }}
               >
                 <option value="">Select a projection...</option>
-                {projectionsList.map((s) => (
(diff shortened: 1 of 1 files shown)


acarbonetto (approved):
Would it be better if the list was sorted by?  
- configs with current project id...
- dashes: "--------"
- the rest...

andy-k-improving:
> Would it be better if the list was sorted by?
> 
> * configs with current project id...
> * dashes: "--------"
> * the rest...


I'd prefer to keep this scoped to the current project for now. 
As users create more projects and projections, a shared list increases the chance of accidentally copying config from the wrong project — and we don't currently have a way to move projections between projects if that happens. 
Filtering by project keeps things clean and avoids silent misconfiguration.

PR #331, opened 2026-07-16
Feat: Skip graph reset for brand new projections

andy-k-improving:
**Description:**

The pipeline now checks for an existing graph before deciding whether to reset.
Brand new graphs are empty by definition — resetting them wastes an API call and polling cycle. 
Existing graphs (retry after failed import) are still reset to clear partial data.

### DIFF
--- a/proxy/src/nx_neptune_proxy/services/pipeline.py
+++ b/proxy/src/nx_neptune_proxy/services/pipeline.py
@@ -18,20 +18,26 @@ async def run_pipeline(projection: Projection) -> None:
     try:
         store.update(projection.id, status="importing")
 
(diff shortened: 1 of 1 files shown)


acarbonetto (commented):
Do we have a place to show the available graphs to the user in the UI?  

That way the user will know if their graph will be reset vs created anew.

andy-k-improving:
> Do we have a place to show the available graphs to the user in the UI?
> 
> That way the user will know if their graph will be reset vs created anew.

Yes, at the moment we have the graph page for me to view all graphs under the same configured prefix.

mykola-pereyma (approved):

PR #332, opened 2026-07-16
Misc: Parallelize graph deletion on project delete

andy-k-improving:
**Description:**

Graph deletions during project delete now run in parallel via `asyncio.gather` instead of sequentially.
Each deletion polls until the graph is gone (~10s intervals), so parallelizing avoids waiting N × timeout for N graphs.
Failures are logged without blocking other deletions.

### DIFF
--- a/proxy/src/nx_neptune_proxy/services/project_deletion.py
+++ b/proxy/src/nx_neptune_proxy/services/project_deletion.py
@@ -23,16 +23,27 @@ async def delete_project(project_id: str) -> None:
     projections = [p for p in projection_store.list() if p.project_id == project_id]
 
     # Delete all graphs in parallel
+    graphs_to_delete = [p for p in projections if p.graph_id]
     results = await asyncio.gather(
-        *[_delete_graph(p.graph_id) for p in projections if p.graph_id],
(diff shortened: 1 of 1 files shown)


proxy/src/nx_neptune_proxy/services/project_deletion.py
```diff
@@ -22,12 +22,16 @@ async def delete_project(project_id: str) -> None:
...
+        if isinstance(r, Exception):
+            logger.warning(f"Failed to delete graph: {r}")
+
     for p in projections:
-        if p.graph_id:
-            try:
-                await _delete_graph(p.graph_id)
-            except Exception as e:
-                logger.warning(f"Failed to delete graph {p.graph_id} for projection {p.id}: {e}")
         projection_store.delete(p.id)
```
acarbonetto: should we keep it around if there's an exception above?

acarbonetto (approved):

proxy/src/nx_neptune_proxy/services/project_deletion.py
andy-k-improving: Yep, updated to perseve the entry, in the case of failed delete.

PR #350, opened 2026-07-23
Add security tests for proxy module

andy-k-improving:
**Description:**

Adds `test_security.py` with 23 tests covering the proxy's security controls:

**Changes**

- Path traversal protection — verifies SPA fallback never serves files outside the UI directory regardless of traversal technique (dot-dot-slash, percent-encoded, backslash, null byte)
- Column allowlist enforcement — verifies the SQLite store rejects updates to columns outside the explicit allowlist
- SQL injection resistance — verifies parameterized queries store injection payloads as literal strings without executing them
- CORS origin restriction — verifies the server does not return permissive CORS headers for disallowed origins
- Credential scanning — verifies the SQLite database contains no secrets or credential patterns after normal operations

**Testing**

All 76 tests pass (53 existing + 23 new):

```
make test 
```


proxy/Dockerfile
```diff
@@ -9,4 +9,4 @@ USER appuser
 EXPOSE 8080
 HEALTHCHECK --interval=10s --timeout=3s --retries=3 \
     CMD wget -qO /dev/null http://localhost:8080/health
-CMD ["uvicorn", "nx_neptune_proxy.app:app", "--host", "0.0.0.0", "--port", "8080"]
+CMD ["uvicorn", "nx_neptune_proxy.app:app", "--host", "127.0.0.1", "--port", "8080"]
```
acarbonetto: Changing the container CMD to bind uvicorn to 127.0.0.1 (loopback only) breaks external access. docker-compose.yml and make run publish -p 8080:8080, but a loopback bind rejects any connection arriving on the published port — the graph-explorer service and host clients get connection-refused. Worse, the HEALTHCHECK hits http://localhost:8080 from inside the container, so it still passes on loopback and masks the outage. Bind to 0.0.0.0 inside the container.

proxy/tests/test_security.py
```diff
@@ -0,0 +1,351 @@
...
+    """SPA fallback must not serve files outside UI_DIR.
+
+    Note: HTTP clients (httpx, browsers) normalize '../' in paths before
+    sending. The security property we verify is: regardless of the path
+    attempted, the response is either 403 or the SPA index.html — never
+    content from outside the UI directory.
+    """
+
+    @pytest.mark.asyncio
+    async def test_dot_dot_slash_serves_spa_not_system_file(self, client):
```
acarbonetto: With ui/ absent, app.py skips registering the spa_fallback route (if UI_DIR.exists()), so test_dot_dot_slash_serves_spa_not_system_file and test_valid_path_returns_200 get 404 and fail. 

This could be confusing.  Consider adding `npm run build` to the proxy_tests job to avoid seeing this error .

proxy/tests/test_security.py
```diff
@@ -0,0 +1,351 @@
...
+    conn.commit()
+    conn.close()
+
+
+# =============================================================================
+# Path traversal attempts on SPA fallback
+# =============================================================================
+
+
+class TestPathTraversal:
```
acarbonetto: Path-traversal tests never hit the server's guard: httpx normalizes ../ client-side, so spa_fallback receives an already-clean path. Deleting the file_path.relative_to(ui_root) guard would not fail these tests.

proxy/tests/test_security.py
```diff
@@ -0,0 +1,351 @@
...
+    async def test_null_byte_traversal(self, client):
+        resp = await client.get("/..%00/etc/passwd")
+        assert resp.status_code in (200, 400, 403)
+        if resp.status_code == 200:
+            assert "<html" in resp.text
+            assert "<title>nx-neptune</title>" in resp.text
+            assert "root:" not in resp.text
+
+    @pytest.mark.asyncio
+    async def test_resolve_prevents_escape(self, client):
```
acarbonetto: re-implements the resolution logic inline instead of calling spa_fallback, so it validates a private copy, not production code

proxy/tests/test_security.py
```diff
@@ -0,0 +1,351 @@
...
+            headers={
+                "Origin": "http://random.com",
+                "Access-Control-Request-Method": "GET",
+            },
+        )
+        allow_origin = resp.headers.get("access-control-allow-origin")
+        assert allow_origin != "*"
+
+    @pytest.mark.asyncio
+    async def test_allowed_origin_gets_cors_headers(self, client):
```
acarbonetto: only checks env parsing. The app builds CORSMiddleware from a module-level settings evaluated at import, so it never exercises the running app's CORS behavior

proxy/tests/test_security.py
```diff
@@ -0,0 +1,351 @@
...
+        retrieved = store.get(p.id)
+        assert retrieved is not None
+
+
+# =============================================================================
+# CORS preflight rejects requests from non-allowed origins
+# =============================================================================
+
+
+class TestCorsRejection:
```
acarbonetto: CORS-rejection tests are near-tautological: they only assert the header ≠ the evil origin, which passes whenever the header is simply absent (or even a permissive wildcard).

proxy/tests/test_security.py
```diff
@@ -0,0 +1,351 @@
...
+        p2 = store.create(database="testdb2", graph_name="test2")
+        assert store.get(p2.id) is not None
+
+    def test_injection_in_sql_query_field(self):
+        payload = "SELECT * FROM t; DROP TABLE projections;--"
+        p = store.create(database="testdb", sql_query=payload, graph_name="test")
+        retrieved = store.get(p.id)
+        assert retrieved.sql_query == payload
+
+    def test_unicode_injection(self):
```
acarbonetto: only asserts the row is not None; it never checks the payload round-trips, so it can't detect NUL truncation or actual injection.

acarbonetto (commented):
A couple of general comments: 
1. Please update the PR name to be more specific
2. Split test_security.py into 5 test files for easier 
3. Consider moving the fixtures to a common file (conftest.py)

proxy/tests/test_security.py
andy-k-improving: Updated tests to use raw ASGI scope so path traversal payloads bypass client-side normalization and actually hit the server guard. Also added an explicit `..` rejection as a first-line control.

proxy/tests/test_security.py
andy-k-improving: Removed the env-parsing-only test — can't test positive CORS without reloading the app (middleware is built at import time).
Kept three rejection tests that hit the running middleware and now assert the header is absent (not just ≠ attacker origin).

proxy/tests/test_security.py
andy-k-improving: Updated the also assert graph name.

proxy/tests/test_security.py
andy-k-improving: Fixed — now asserts the header is absent (is None) rather than just ≠ the attacker origin.

proxy/tests/test_security.py
andy-k-improving: Updated `make` target to always build the `ui`, and in the case the `ui` indeed missing during the test time, warning message will be shown.
```
================================== short test summary info ===================================
SKIPPED [1] tests/test_security.py:90: ui/ directory not built — run 'make ui' first
SKIPPED [1] tests/test_security.py:96: ui/ directory not built — run 'make ui' first
SKIPPED [1] tests/test_security.py:102: ui/ directory not built — run 'make ui' first
SKIPPED [1] tests/test_security.py:108: ui/ directory not built — run 'make ui' first
SKIPPED [1] tests/test_security.py:114: ui/ directory not built — run 'make ui' first
SKIPPED [1] tests/test_security.py:124: ui/ directory not built — run 'make ui' first
SKIPPED [1] tests/test_security.py:141: ui/ directory not built — run 'make ui' first
========================= 68 passed, 7 skipped, 2 warnings in 0.53s =====================
```

andy-k-improving:
> A couple of general comments:
> 
> 1. Please update the PR name to be more specific
> 2. Split test_security.py into 5 test files for easier
> 3. Consider moving the fixtures to a common file (conftest.py)

Done.

proxy/pyproject.toml
```diff
@@ -61,3 +61,4 @@ exclude = []
 
 [tool.pytest.ini_options]
 asyncio_mode = "auto"
+addopts = "-rs"
```
acarbonetto, line 64: addopts is correct, not `adopts`?

acarbonetto (approved):

proxy/pyproject.toml
andy-k-improving: Yes, it's addopts.
https://docs.pytest.org/en/stable/reference/reference.html#confval-addopts

PR #361, opened 2026-07-27
Add CSRF protection, prefix-scoped deletes, and error sanitization to proxy

andy-k-improving:
**Description:**

Security hardening for the proxy module.

**Changes**

- **CSRF protection** — Require `X-Requested-With` header on all state-changing requests (POST/PUT/DELETE). This forces browser preflight on cross-origin requests, preventing drive-by CSRF from malicious tabs. Frontend updated to send the header on all API calls.

- **Prefix-scoped deletes** — Graph delete operations now verify the target graph name starts with the configured prefix (`nxp-`). Prevents accidental or malicious deletion of graphs not created by this tool.

- **Error message sanitization** — AWS error responses are stripped of ARNs, account IDs, and access keys before being returned to the client or persisted to SQLite.

- **Docker loopback binding** — Published ports in docker-compose.yml bound to `127.0.0.1`, matching the standalone deployment security posture.

- **react-router bump** — Upgraded from 7.16.0 to 8.3.0, resolving known high-severity vulnerabilities.

**Testing**

89 tests pass. New test files:
- `test_csrf.py` — 7 tests (rejected without header, allowed with header, safe methods unaffected)
- `test_error_sanitization.py` — 7 tests (ARN/account ID/key stripping, normal messages preserved)
- Frontend builds clean with `tsc --noEmit && vite build`

### DIFF
--- a/proxy/src/nx_neptune_proxy/routers/graph.py
+++ b/proxy/src/nx_neptune_proxy/routers/graph.py
@@ -8,7 +8,7 @@ from __future__ import annotations
 import logging
 
 from fastapi import APIRouter, BackgroundTasks, HTTPException
-
+from nx_neptune_proxy.config import Settings
 from nx_neptune.clients.client_factory import ClientFactory
 from nx_neptune_proxy.services.graph_state_machine import (
     InvalidTransitionError,
@@ -52,6 +52,18 @@ def perform_action(graph_id: str, action: str, background_tasks: BackgroundTasks
     if not current_status:
         logger.warning(f"Graph {graph_id} returned empty status: {resp}")
         raise HTTPException(status_code=502, detail="Graph returned no status")
+
+    # Prefix guard: only allow destructive actions on graphs managed by this tool
+    if action == "delete":
+
+
+        prefix = Settings.from_env().graph_prefix
+        graph_name = resp.get("name", "")
+        if not graph_name.startswith(prefix):
+            raise HTTPException(
+                status_code=403,
+                detail=f"Graph '{graph_name}' is not managed by this tool (missing '{prefix}' prefix)",
+            )
 
     try:
         # Validate early (execute_transition also validates, but fail fast for the user)
--- a/proxy/tests/test_app.py
+++ b/proxy/tests/test_app.py
@@ -10,7 +10,7 @@ from nx_neptune_proxy.app import app
 @pytest.fixture
 def client():
     transport = ASGITransport(app=app)
-    return AsyncClient(transport=transport, base_url="http://test")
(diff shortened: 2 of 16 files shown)


proxy/src/nx_neptune_proxy/routers/graph.py
```diff
@@ -53,6 +53,18 @@ def perform_action(graph_id: str, action: str, background_tasks: BackgroundTasks
         logger.warning(f"Graph {graph_id} returned empty status: {resp}")
         raise HTTPException(status_code=502, detail="Graph returned no status")
 
+    # Prefix guard: only allow destructive actions on graphs managed by this tool
+    if action == "delete":
+
+
+        prefix = Settings.from_env().graph_prefix
+        graph_name = resp.get("name", "")
```
acarbonetto, line 61: We should tag resources created by our tool instead of going in this direction.  
Ideally, we have a tag with the user that created it too... then we can check both tags: 
1. created by this user
2. created by Graph Studio
What do you think?

proxy/tests/test_app.py
```diff
@@ -10,7 +10,7 @@
 @pytest.fixture
 def client():
     transport = ASGITransport(app=app)
-    return AsyncClient(transport=transport, base_url="http://test")
+    return AsyncClient(transport=transport, base_url="http://test", headers={"X-Requested-With": "nx-neptune"})
```
acarbonetto, line 13: Please create a follow-up ticket and upgrade this with session tokens.  One strategy we could use, is deploy nx-neptune and graph studio with a unique (random of not provided) session uuid, and verify that the session tokens match.  This would secure a one-to-one relationship between graph studio and nx-neptune (or deployed to multiple graph studio with the same token).  

We can do even better with the a refreshable token, or even a token generation mechanism to secure the relationship.

acarbonetto: That being said, we don't yet have a use case design to use multiple nx-neptunes with one graph studio... so that could be a limitation.  I'm fine with that.

proxy/src/nx_neptune_proxy/routers/graph.py
andy-k-improving: This won't be a trivial as the session manager  is deigned with prefix, we will need to work on this separately.

proxy/tests/test_app.py
andy-k-improving: Follow up ticket created.

proxy/src/nx_neptune_proxy/routers/graph.py
andy-k-improving: Follow up ticket created.

PR #366, opened 2026-07-29
Fix: Bind Docker published ports to 127.0.0.1

andy-k-improving:
Summary
  
Bind all published ports in docker-compose.yml to 127.0.0.1, preventing remote access to development containers.
Matches the standalone deployment security posture.
  
  Changes
  
  - proxy/docker-compose.yml — 8080:8080 → 127.0.0.1:8080:8080, same for ports 443 and 80

### DIFF
--- a/proxy/docker-compose.yml
+++ b/proxy/docker-compose.yml
@@ -4,7 +4,7 @@ services:
       context: ..
       dockerfile: proxy/Dockerfile
     ports:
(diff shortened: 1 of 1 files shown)


acarbonetto (approved):
Seems fine.  We should bind to an environment variable to deployable config in the future if that's desired.

PR #368, opened 2026-07-29
Fix: Add CSRF protection via X-Requested-With header

andy-k-improving:
Summary

Require X-Requested-With header on all state-changing requests (POST/PUT/DELETE). This forces browser preflight on cross-origin requests, preventing drive-by CSRF from malicious tabs. 
Frontend updated to send the header on all API calls.

Changes

- proxy/src/nx_neptune_proxy/app.py — CSRF middleware + X-Requested-With in CORS allow_headers
- proxy/ui-src/src/api/index.ts — send header on all fetch calls
- proxy/tests/conftest.py — client fixture sends header; new bare_client without it
- proxy/tests/test_csrf.py — 7 tests covering blocked/allowed/safe-method cases
- proxy/tests/test_app.py, test_graph.py, test_metadata.py, test_projection.py — fixtures
updated

Testing

75 passed, 7 skipped (UI not built)

### DIFF
--- a/proxy/src/nx_neptune_proxy/app.py
+++ b/proxy/src/nx_neptune_proxy/app.py
@@ -16,6 +16,7 @@ from fastapi.staticfiles import StaticFiles
 from nx_neptune_proxy.config import Settings
 from nx_neptune_proxy.routers.graph import router as graph_router
 from nx_neptune_proxy.routers.metadata import router as metadata_router
+from nx_neptune_proxy.utils.sanitize import sanitize_error_message
 
 from nx_neptune_proxy.routers.projection import router as projection_router
 from nx_neptune_proxy.routers.project import router as project_router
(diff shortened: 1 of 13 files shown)


proxy/src/nx_neptune_proxy/app.py
```diff
@@ -46,10 +46,31 @@
...
     allow_origins=settings.allowed_origins,
     allow_methods=["GET", "POST", "PUT", "DELETE"],
-    allow_headers=["Content-Type", "Authorization"],
+    allow_headers=["Content-Type", "Authorization", "X-Requested-With"],
 )
 
 
+# --- CSRF protection middleware ---
+
+_CSRF_SAFE_METHODS = {"GET", "HEAD", "OPTIONS"}
```
acarbonetto, line 56: not sure I understand why we wouldn't do this for read-only calls too.

acarbonetto (approved):

proxy/src/nx_neptune_proxy/app.py
acarbonetto: oh, nevermind.  we could be extra cautious and check all calls, but I see only state-changing calls are prone to malicious CRSF.

PR #369, opened 2026-07-29
Fix: Add prefix-scoped delete guard

andy-k-improving:
Summary

Graph delete operations now verify the target graph name starts with the configured prefix (nxp-). Prevents accidental or malicious deletion of graphs not created by this tool.

Changes

- proxy/src/nx_neptune_proxy/routers/graph.py — prefix check before delete action
- proxy/src/nx_neptune_proxy/routers/metadata.py — prefix check on delete_neptune_graph
- proxy/src/nx_neptune_proxy/routers/projection.py — prefix check on delete_projection_graph
- proxy/tests/test_graph.py — mock response includes name: "nxp-test-graph"

### DIFF
--- a/proxy/src/nx_neptune_proxy/routers/graph.py
+++ b/proxy/src/nx_neptune_proxy/routers/graph.py
@@ -8,7 +8,7 @@ from __future__ import annotations
 import logging
 
 from fastapi import APIRouter, BackgroundTasks, HTTPException
-
+from nx_neptune_proxy.config import Settings
 from nx_neptune.clients.client_factory import ClientFactory
 from nx_neptune_proxy.services.graph_state_machine import (
     InvalidTransitionError,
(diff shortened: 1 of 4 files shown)


proxy/src/nx_neptune_proxy/routers/graph.py
```diff
@@ -53,6 +53,16 @@ def perform_action(graph_id: str, action: str, background_tasks: BackgroundTasks
         logger.warning(f"Graph {graph_id} returned empty status: {resp}")
         raise HTTPException(status_code=502, detail="Graph returned no status")
 
+    # Prefix guard: only allow destructive actions on graphs managed by this tool
+    if action == "delete":
+        prefix = Settings.from_env().graph_prefix
+        graph_name = resp.get("name", "")
+        if not graph_name.startswith(prefix):
```
acarbonetto, line 60: we should be using resource tagging to check this, rather than a prefix (which is not very safe). 
consider tagging all graphs created with a session name and origin app (nx-neptune) and check for the presence of both before deleting.

acarbonetto (commented):
I'm okay with this, but I'd much prefer that we tag our resources instead of relying on the name of the graph. I understand that's the original use-case, but tagging is much stronger and less prone to mistakes.

proxy/src/nx_neptune_proxy/routers/graph.py
andy-k-improving: Agreed — tagging is the right long-term approach. The prefix check is a stopgap since the session manager currently relies on it for all lifecycle ops.

PR #379, opened 2026-07-31
Pass explicit limit=10 to preview endpoint

andy-k-improving:
**Description:**

The UI was relying on the server default for preview row count.
Make it explicit so the intent is clear and the behavior doesn't silently change if the server default is modified.


acarbonetto (approved):
I think this is already defaulted to 10

  The change is functionally a no-op. Both layers already default to 10:
  - API client: preview: (id: string, limit = 10) => … (api/index.ts:84)
  - Backend handler: preview_projection(projection_id, limit: int = Query(10, …))

PR #380, opened 2026-07-31
Fix: Add prefix-scoped delete guard

andy-k-improving:
Summary

Graph delete operations now verify the target graph name starts with the configured prefix (nxp-). Prevents accidental or malicious deletion of graphs not created by this tool.

Changes

- proxy/src/nx_neptune_proxy/routers/graph.py — prefix check before delete action
- proxy/src/nx_neptune_proxy/routers/metadata.py — prefix check on delete_neptune_graph
- proxy/src/nx_neptune_proxy/routers/projection.py — prefix check on delete_projection_graph
- proxy/tests/test_graph.py — mock response includes name: "nxp-test-graph"


andy-k-improving:
See: https://github.com/awslabs/nx-neptune/pull/369

proxy/src/nx_neptune_proxy/routers/projection.py
```diff
@@ -167,6 +167,15 @@ def delete_projection_graph(projection_id: str, background_tasks: BackgroundTask
         raise HTTPException(status_code=409, detail="No graph associated with this projection")
     if p.status == "deleting":
         raise HTTPException(status_code=409, detail="Already deleting")
+
+    # Prefix guard: only delete graphs managed by this tool
+    prefix = Settings.from_env().graph_prefix
```
acarbonetto: The guard checks `p.graph_name.startswith(prefix)`, but `p.graph_name` is the bare, user-supplied name stored in the DB. The real Neptune graph is created as `f"{graph_prefix}{projection.graph_name}"` (pipeline.py:16). So a projection with `graph_name="mydata"` owns a graph named nxp-mydata, but the guard tests `"mydata".startswith("nxp-")` → False → HTTP 403. The delete-graph feature is broken for every normally-named projection; only a projection the user happened to name nxp-… would pass.

acarbonetto: the sibling guards in graph.py:59 and metadata.py:96 read the authoritative name from client.get_graph(...)'s response (resp.get("name")), which is the prefixed name. Only the projection.py copy reads the stale DB column

proxy/src/nx_neptune_proxy/routers/projection.py
```diff
@@ -167,6 +167,15 @@ def delete_projection_graph(projection_id: str, background_tasks: BackgroundTask
...
+    if p.graph_name and not p.graph_name.startswith(prefix):
```
acarbonetto: p.graph_name=None skips this and will delete it all the time. 
...which is technically allowed in ProjectionCreate

proxy/src/nx_neptune_proxy/routers/metadata.py
```diff
@@ -90,6 +90,15 @@ def list_neptune_graphs():
 def delete_neptune_graph(graph_id: str):
     """Delete a Neptune Analytics graph"""
     client = ClientFactory().neptune()
+    # Verify graph belongs to this tool (prefix guard)
+    prefix = Settings.from_env().graph_prefix
```
acarbonetto: If the graph was already deleted or never existed, boto3 raises ResourceNotFoundException, which now surfaces as a 500/502

acarbonetto (approved):
graph.py:57, metadata.py:96, projection.py:172 - The same Settings.from_env().graph_prefix + startswith + 403 with an identical detail string is pasted in three routers
Can we consolidate?  

Otherwise LGMT

proxy/src/nx_neptune_proxy/routers/projection.py
andy-k-improving: Done.

proxy/src/nx_neptune_proxy/routers/projection.py
andy-k-improving: Done.

PR #383, opened 2026-08-05
Feat: Add S3 storage backend for project import/export

andy-k-improving:
**Description:**

Extends project import/export with optional S3 as a storage backend. When `NX_NEPTUNE_EXPORT_BUCKET` is configured, users can save exports to S3 and load from S3 directly in the UI.

## Changes

**Backend**
- 3 new endpoints: `POST /{id}/export/s3`, `GET /import/s3/list`, `POST /import/s3`
- Exports tagged with `graph_studio=true`; list endpoint returns last 10 `.json` files sorted by date
- Duplicate key check (409) before upload
- Friendly error messages for S3 failures
- Config: `export_bucket` field from `NX_NEPTUNE_EXPORT_BUCKET` env var, exposed via `/metadata/config`

**Frontend**
- Export button → dropdown ("Download" | "Save to S3") when bucket is configured
- Import button → dropdown ("Upload file" | "Load from S3") when bucket is configured
- `S3ImportDialog` modal with file list table, click-to-select, confirmation before import

https://github.com/user-attachments/assets/93e64737-ecc8-4ff8-8b57-58f45aad2e3d

https://github.com/user-attachments/assets/63cd657b-500f-440b-a5e4-3e4bdbc2699d

## Testing
- 20 new S3 tests, 120 total passing

## Configuration
Set `NX_NEPTUNE_EXPORT_BUCKET=bucket-name` or `NX_NEPTUNE_EXPORT_BUCKET=bucket-name/prefix` to enable.

### DIFF
--- a/nx_neptune/clients/iam_client.py
+++ b/nx_neptune/clients/iam_client.py
@@ -701,16 +701,18 @@ class IamClientWrapper:
         return results
 
 
-def split_s3_arn_to_bucket_and_path(s3_arn: str) -> tuple:
+def split_s3_arn_to_bucket_and_path(s3_arn: str, include_path: bool = True) -> tuple:
     """
     Splits out the s3 arn as a bucket and path
-    :param s3_arn:
+    :param s3_arn: S3 URI (e.g., 's3://bucket-name/prefix/') or plain path (e.g., 'bucket-name/prefix')
+    :param include_path: If True (default), returns (bucket, path). If False, returns (bucket, "").
     :return: (str, str) - a tuple containing the bucket and path
     """
 
-    bucket_path = s3_arn.replace("s3://", "").split("/")
+    raw = s3_arn.replace("s3://", "") if s3_arn.startswith("s3://") else s3_arn
+    bucket_path = raw.split("/")
     bucket_name = bucket_path.pop(0)
-    bucket_path_str = "/".join(bucket_path)
+    bucket_path_str = "/".join(bucket_path) if include_path else ""
 
     return bucket_name, bucket_path_str
 
--- a/nx_neptune/clients/iam_client.py
+++ b/nx_neptune/clients/iam_client.py
@@ -701,18 +701,17 @@ class IamClientWrapper:
         return results
 
 
-def split_s3_arn_to_bucket_and_path(s3_arn: str, include_path: bool = True) -> tuple:
+def split_s3_arn_to_bucket_and_path(s3_arn: str) -> tuple:
     """
     Splits out the s3 arn as a bucket and path
     :param s3_arn: S3 URI (e.g., 's3://bucket-name/prefix/') or plain path (e.g., 'bucket-name/prefix')
-    :param include_path: If True (default), returns (bucket, path). If False, returns (bucket, "").
     :return: (str, str) - a tuple containing the bucket and path
     """
 
     raw = s3_arn.replace("s3://", "") if s3_arn.startswith("s3://") else s3_arn
     bucket_path = raw.split("/")
     bucket_name = bucket_path.pop(0)
-    bucket_path_str = "/".join(bucket_path) if include_path else ""
+    bucket_path_str = "/".join(bucket_path)
 
     return bucket_name, bucket_path_str
 
--- a/proxy/src/nx_neptune_proxy/routers/metadata.py
+++ b/proxy/src/nx_neptune_proxy/routers/metadata.py
@@ -26,7 +26,7 @@ def get_config():
     return {
         "region": settings.region or "",
         "graph_prefix": settings.graph_prefix,
-        "export_bucket": settings.export_bucket or None,
+        "export_bucket": settings.export_bucket or "",
     }
 
 
--- a/proxy/src/nx_neptune_proxy/routers/metadata.py
+++ b/proxy/src/nx_neptune_proxy/routers/metadata.py
@@ -26,7 +26,7 @@ def get_config():
     return {
         "region": settings.region or "",
         "graph_prefix": settings.graph_prefix,
-        "export_bucket": settings.export_bucket or "",
+        "config_bucket": settings.config_bucket or "",
     }
 
 
--- a/proxy/src/nx_neptune_proxy/routers/metadata.py
+++ b/proxy/src/nx_neptune_proxy/routers/metadata.py
@@ -23,7 +23,11 @@ router = APIRouter(prefix="/api/v0/metadata", tags=["metadata"])
 def get_config():
     """Get server-side configuration"""
     settings = Settings.from_env()
-    return {"region": settings.region or "", "graph_prefix": settings.graph_prefix}
+    return {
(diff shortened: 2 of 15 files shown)


proxy/src/nx_neptune_proxy/routers/metadata.py
```diff
@@ -22,7 +22,11 @@
 def get_config():
     """Get server-side configuration"""
     settings = Settings.from_env()
-    return {"region": settings.region or "", "graph_prefix": settings.graph_prefix}
+    return {
+        "region": settings.region or "",
+        "graph_prefix": settings.graph_prefix,
+        "export_bucket": settings.export_bucket or None,
```
acarbonetto: we're using two kinds of string defaults.  Can we use "" in both cases?  Or None is both cases?

proxy/src/nx_neptune_proxy/utils/aws_helper.py
```diff
@@ -30,3 +30,67 @@ def unpack_query_results(rows: list) -> dict:
...
+    Args:
+        s3: boto3 S3 client
+        bucket: S3 bucket name
+        prefix: Optional key prefix (without trailing slash)
+        max_keys: Maximum number of objects to fetch from S3
+
+    Returns:
+        List of S3 object dicts (Key, LastModified, etc.) filtered to .json files only.
+    """
+    list_kwargs = {"Bucket": bucket, "MaxKeys": max_keys}
```
acarbonetto: A single un-paginated list_objects_v2 call with MaxKeys=100, then sorts by LastModified. But S3 returns keys in lexicographic order, not by date — so once the prefix holds >100 objects, the sort only sees the first 100 alphabetical keys. Recently exported projects whose names sort late never appear in the "last 10 exports" list, and can't be imported. Worse: if the first 100 lexicographic keys are all non-.json, the list comes back empty despite exports existing.
  This breaks the feature at scale.

proxy/src/nx_neptune_proxy/routers/project_io.py
```diff
@@ -5,120 +5,178 @@
...
-        )
+        return {
+            "files": [_s3_object_to_entry(obj) for obj in recent]
+        }
+    except ClientError as e:
+        raise HTTPException(status_code=502, detail=friendly_s3_error(e))
+
+
+@router.post("/import/s3", summary="Import a project from S3", status_code=201)
+def import_project_from_s3(request_body: dict):
```
acarbonetto, line 182: The key from the request is passed straight to get_object against the configured bucket, with no validation that it lies under the configured prefix. On a shared bucket (s3://shared-bucket/exports), a client can POST {"key": "secrets/other-team-data.json"} and read/import any object anywhere in the bucket. The listing path filters by prefix; the import path does not. Path-scoping bypass.

nx_neptune/clients/iam_client.py
```diff
@@ -701,16 +701,18 @@ def validate_permissions(
...
+    :param include_path: If True (default), returns (bucket, path). If False, returns (bucket, "").
     :return: (str, str) - a tuple containing the bucket and path
     """
 
-    bucket_path = s3_arn.replace("s3://", "").split("/")
+    raw = s3_arn.replace("s3://", "") if s3_arn.startswith("s3://") else s3_arn
+    bucket_path = raw.split("/")
     bucket_name = bucket_path.pop(0)
-    bucket_path_str = "/".join(bucket_path)
+    bucket_path_str = "/".join(bucket_path) if include_path else ""
```
acarbonetto: If NX_NEPTUNE_EXPORT_BUCKET=s3://my-bucket/exports/, split_s3_arn_to_bucket_and_path returns prefix='exports/', and _build_s3_key builds f"{prefix}/{filename}" → exports//myproj_….json. New objects get an empty path segment, and any pre-existing exports under exports/ (single slash) won't match the exports// list prefix — so they vanish from the import list.

proxy/src/nx_neptune_proxy/routers/project_io.py
```diff
@@ -5,120 +5,178 @@
...
+        raise HTTPException(status_code=404, detail="Project not found")
+
+    projections = projection_store.list_by_project(project_id)
+    payload = ProjectExportPayload.from_project(p, projections)
+    return payload.model_dump(), p.name
+
 
+def _build_s3_key(name: str, prefix: str) -> tuple[str, str]:
+    """Build S3 key from project name and prefix. Returns (key, filename)."""
+    sanitized = sanitize_s3_key_name(name)
```
acarbonetto: ensure that sanitized is non-empty

proxy/ui-src/src/pages/Projections.tsx
```diff
@@ -1,8 +1,8 @@
-import { useEffect, useState } from "react";
+import { useEffect, useState, useRef } from "react";
 import { useSearchParams, NavLink } from "react-router";
 import { projection, metadata, projectApi, graphActions, type Projection, type Project, type Inflight } from "../api";
 import { Card, Button, RefreshButton } from "../components/ui";
-import { X, ExternalLink, Trash2, Square, Play, AlertTriangle, ChevronRight, ChevronDown, Download } from "lucide-react";
+import { X, ExternalLink, Trash2, Square, Play, AlertTriangle, ChevronRight, ChevronDown, Download, ChevronUp } from "lucide-react";
```
acarbonetto: ChevronUp not used

nx_neptune/clients/iam_client.py
```diff
@@ -701,16 +701,18 @@ def validate_permissions(
         return results
 
 
-def split_s3_arn_to_bucket_and_path(s3_arn: str) -> tuple:
+def split_s3_arn_to_bucket_and_path(s3_arn: str, include_path: bool = True) -> tuple:
```
acarbonetto: include_path never set to `false`

acarbonetto (approved):
Please take a look at the comments.  Please comment limitations if necessary

proxy/src/nx_neptune_proxy/routers/metadata.py
andy-k-improving: " "

proxy/src/nx_neptune_proxy/utils/aws_helper.py
andy-k-improving: Fixed by paginating through all objects under the prefix, filtering to `.json`, sorting by `LastModified` descending in Python, and returning the top 10.

nx_neptune/clients/iam_client.py
andy-k-improving: Normalized the prefix with `rstrip("/")` in `_get_export_bucket_tuple` so `s3://bucket/exports/` produces prefix `"exports"` instead of `"exports/"`.

nx_neptune/clients/iam_client.py
andy-k-improving: Removed.

PR #399, opened 2026-08-13
Fix: Make project_id mandatory for projections

andy-k-improving:
## Summary

Projections can no longer exist without a parent project. Enforced at API, DB, and UI levels.

Screenshot 2026-08-13 at 3 32 50 PM

### Backend
- `ProjectionCreate.project_id` — changed from `Optional[str] = None` to `str` (required). Pydantic returns 422 if missing.
- DB schema: `project_id TEXT NOT NULL` with foreign key constraint on projections table.

### Frontend
- Project dropdown is now read-only (disabled) — project is set by navigation context, not user selection.
- Removed the `+` create project button from Import page.
- `loadProjection` now sets `projectId` from the loaded projection's `project_id` field.
- `ensureProjection()` throws if no project is selected.
- TypeScript: `project_id` is non-optional in `Projection` interface.

### Tests
- Added `test_project_id` fixture in `conftest.py` for tests that call `store.create()` directly.
- Updated `test_sql_injection.py`, `test_column_allowlist.py`, `test_no_secrets.py`, `test_projection.py` to always pass `project_id`.

### DIFF
--- a/proxy/src/nx_neptune_proxy/routers/schemas.py
+++ b/proxy/src/nx_neptune_proxy/routers/schemas.py
@@ -61,7 +61,7 @@ class ProjectionCreate(BaseModel):
     graph_name: Optional[str] = None
     graph_memory_gb: int = 16
     s3_staging_bucket: Optional[str] = None
-    project_id: Optional[str] = None
+    project_id: str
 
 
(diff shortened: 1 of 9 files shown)


acarbonetto (approved):
lol

PR #406, opened 2026-08-17
Feat: Restrict proxy API to local-only requests

andy-k-improving:
**Summary**

Adds request validation to the proxy API to ensure it only accepts connections from the local machine, and requires a per-run access token on all API routes. Includes a small frontend change so the bundled UI picks up the token automatically.

**Motivation**

The proxy API previously accepted requests from any origin without validating the host header, and had no authentication on any route — any local process, or a browser tricked via DNS rebinding, could call it directly. This tightens the server to only respond to legitimate local requests and requires callers to present an access token, matching the intended usage pattern.

**Changes**

Five controls added:

| Control | Purpose |
|---------|---------|
| TrustedHostMiddleware | Ensures requests are addressed to localhost/127.0.0.1/[::1] |
| Origin validation middleware | Confirms the request originates from a local source |
| Loopback enforcement | Prevents the server from binding to a non-loopback address |
| Disable /docs | Removes the auto-generated API schema endpoint |
| Per-run access token | Requires `Authorization: Bearer <token>` on all `/api/*` routes; a fresh token is generated in memory each time the proxy starts. `/health` and `/api/v0/info` remain open. The bundled UI receives the token automatically (injected into the served page); other callers can read it from the proxy's startup output. |


proxy/src/nx_neptune_proxy/config.py
```diff
@@ -27,6 +28,17 @@ def from_env(cls) -> "Settings":
...
             allowed_origins=origins,
             port=int(os.environ.get("PORT", "8080")),
+            host=os.environ.get("HOST", "127.0.0.1"),
             region=os.environ.get("AWS_DEFAULT_REGION", os.environ.get("AWS_REGION", "")),
             graph_prefix=os.environ.get("GRAPH_PREFIX", "nxp-"),
         )
+
+    def validate_host(self) -> None:
+        """Refuse to start on non-loopback address."""
+        loopback = {"127.0.0.1", "::1", "localhost"}
```
acarbonetto, line 38: ## Loopback enforcement is disconnected from the real bind

`validate_host()` checks the HOST env var, but the actual uvicorn bind address is set independently. The Dockerfile CMD hardcodes `uvicorn ... --host 0.0.0.0` and never sets HOST, so `validate_host()` see the default `127.0.0.1`, passes, yet the server actually binds to all interfaces. The host field is only read by validate_host and never passed to uvicorn — so the check gives false assurance. Only the docker `127.0.0.1:8080:8080` port mapping limits exposure, not this code.

Fix: bind uvicorn to `settings.host` (or change the Dockerfile CMD to bind loopback / read HOST).

proxy/src/nx_neptune_proxy/app.py
```diff
@@ -51,6 +60,34 @@
 )
 
 
+# --- Origin validation middleware (server-side, defense-in-depth) ---
+
+_ALLOWED_ORIGIN_HOSTS = {"localhost", "127.0.0.1", "[::1]"}
```
acarbonetto, line 76: ## IPv6 loopback origins wrongly rejected

`_ALLOWED_ORIGIN_HOSTS` contains the bracketed `'[::1]'`, but `urlparse(origin).hostname` returns the unbracketed `'::1'`. A legitimate `Origin: http://[::1]:8080` request gets a 403 `origin_rejected`. (Verified: `urlparse('http://[::1]:8080').hostname == '::1'`.)

proxy/src/nx_neptune_proxy/app.py
```diff
@@ -41,7 +45,12 @@
 
 # --- App ---
 
-app = FastAPI(title="nx-neptune-proxy", version="0.1.0", docs_url="/docs", redoc_url=None)
+app = FastAPI(title="nx-neptune-proxy", version="0.1.0", docs_url=None, redoc_url=None)
+
+# --- TrustedHost middleware (blocks DNS rebinding) ---
+
+
+app.add_middleware(TrustedHostMiddleware, allowed_hosts=["localhost", "127.0.0.1", "[::1]"])
```
acarbonetto: ## TrustedHostMiddleware [::1] entry can never match

Starlette computes host via `host.split(':')[0]`, so an IPv6 `Host: [::1]:8080` yields `'['`, never matching `'[::1]'`. All IPv6 loopback requests get a 400 Invalid host header. The `'[::1]'` entry is dead code and IPv6 loopback is unsupported despite appearing allowed.

proxy/src/nx_neptune_proxy/config.py
```diff
@@ -27,6 +28,17 @@ def from_env(cls) -> "Settings":
             log_level=os.environ.get("LOG_LEVEL", "INFO").upper(),
             allowed_origins=origins,
             port=int(os.environ.get("PORT", "8080")),
+            host=os.environ.get("HOST", "127.0.0.1"),
```
acarbonetto, line 31: ## HOST setting is dead configuration

  The new HOST setting is parsed into Settings but never configures the actual server bind. A user setting `HOST=0.0.0.0` hits `SystemExit`, while a user relying on HOST to select the interface gets no effect — uvicorn's --host is what binds. Misleading knob.

proxy/tests/test_dns_rebinding.py
```diff
@@ -0,0 +1,104 @@
...
+
+class TestDocsDisabled:
+    """API docs endpoints must not be accessible."""
+
+    @pytest.mark.asyncio
+    async def test_docs_not_found(self):
+        transport = ASGITransport(app=app)
+        client = AsyncClient(transport=transport, base_url="http://localhost")
+        resp = await client.get("/docs")
+        # Should not return 200 with Swagger UI
```
acarbonetto, line 96: ## Near-tautological docs-disabled tests

`test_docs_not_found` / `test_redoc_not_found` pass even if docs were re-enabled or the SPA fallback served content. With `docs_url=None`, /docs falls through to the SPA catch-all returning index.html (200); the assertion passes only because index.html lacks 'swagger'. Provides little protection for the control it claims to cover.

acarbonetto (changes requested):

proxy/src/nx_neptune_proxy/app.py
```diff
@@ -38,10 +43,25 @@
 )
 logger = logging.getLogger("nx_neptune_proxy")
 
+# --- Proxy access token (per-run bearer token, distinct from AWS credentials) ---
+# Printed directly to stdout (not through the structured logger) so it never
+# ends up in aggregated/forwarded logs. Only a caller with access to this
+# process's own console output, or the bundled UI (which receives it via
+# index.html), can read it.
+print(f"\nProxy access token for this run: {get_token()}\n")
```
acarbonetto: The comment claims stdout "never ends up in aggregated/forwarded logs," but the shipped Dockerfile runs the app as PID 1, so this print goes straight to container stdout — exactly what docker logs, log drivers, and CloudWatch capture. In any containerized run (make up), the per-run bearer token is written verbatim into forwarded logs, defeating the "in memory only, never logged in full" rationale.

proxy/src/nx_neptune_proxy/config.py
andy-k-improving: Fixed. `HOST` was never passed to uvicorn, so `validate_host()` checked a disconnected default.

Dockerfile `CMD` now reads `${HOST}`/`${PORT}` for real, and `ALLOWED_HOSTS` lets it explicitly opt in to `0.0.0.0` (additive to the loopback default). Verified with real container runs — default works, bad `HOST` now fails to start. Added `test_config.py`.

andy-k-improving:
Create multiple PRs to address each point in specific, closing this for now.

PR #407, opened 2026-08-17
Fix: Guard stop/start with managed-graph prefix check

andy-k-improving:
**Summary**

Extends the `nxp-` prefix guard to all graph modification endpoints. Previously only delete was guarded — stop and start accepted any graph ID in the account.

**Motivation**

Several endpoints didn't apply the prefix check, so they operated on graphs outside the set this tool manages:

- `POST /graphs/{id}/stop` and `POST /graphs/{id}/start` could stop/start any Neptune graph in the account. Both dispatch through `perform_action` in `graph.py`, where the guard was scoped to `if action == "delete"` — so stop/start slipped past it.
- `GET /graphs/{id}/actions` reported status for arbitrary graphs
- An empty `GRAPH_PREFIX` made the guard a no-op (`"".startswith("") == True`)

Note: graph deletion (`delete_neptune_graph` in `metadata.py`) was already guarded before this PR — it's a separate handler and not part of the `perform_action` dispatch.

**Changes**

- Moved `assert_managed_graph` above the action dispatch in `perform_action` — now covers stop/start/delete
- Added the prefix guard to `get_available_actions`
- Added `Settings.validate()` that rejects an empty `GRAPH_PREFIX` at startup
- Generalized the error message from "Refusing to delete" to "Refusing to operate on"

**Intentionally not guarded**

- `get_graph_summary` — read-only (just node/edge counts), and guarding it adds an extra `get_graph` call per request.
- `get_inflight_status` — reads a local in-memory map that is only populated by guarded actions, so it only ever contains managed graphs. Unmanaged IDs return null. No Neptune call.

### DIFF
--- a/proxy/src/nx_neptune_proxy/config.py
+++ b/proxy/src/nx_neptune_proxy/config.py
@@ -67,3 +67,26 @@ class Settings:
                 "An empty prefix disables the managed-graph safety guard, "
                 "allowing operations on any graph in the account."
             )
+
+
+# Single validated Settings instance shared across the process. Built and
+# validated once on first access, then reused, so the value the startup
+# guard validates is the exact same instance every consumer (e.g. the
+# managed-graph guard) reads — not an independent, unvalidated re-read.
+_settings: "Settings | None" = None
+
+
+def get_settings() -> "Settings":
+    """Return the process-wide validated Settings instance.
+
+    Loads from the environment and validates on first call, then caches.
+    All runtime consumers should read configuration through this, rather
+    than calling ``Settings.from_env()`` directly, so everyone shares the
+    one instance that startup validation checked.
+    """
+    global _settings
+    if _settings is None:
+        settings = Settings.from_env()
+        settings.validate()
+        _settings = settings
+    return _settings
--- a/proxy/src/nx_neptune_proxy/config.py
+++ b/proxy/src/nx_neptune_proxy/config.py
@@ -58,3 +58,12 @@ class Settings:
             ),
             graph_prefix=os.environ.get("GRAPH_PREFIX", "nxp-"),
         )
+
+    def validate(self) -> None:
+        """Validate settings at startup. Raises SystemExit on invalid config."""
+        if not self.graph_prefix:
+            raise SystemExit(
+                "ERROR: GRAPH_PREFIX must not be empty. "
+                "An empty prefix disables the managed-graph safety guard, "
+                "allowing operations on any graph in the account."
+            )
--- a/proxy/src/nx_neptune_proxy/config.py
+++ b/proxy/src/nx_neptune_proxy/config.py
@@ -60,12 +60,24 @@ class Settings:
         )
 
     def validate(self) -> None:
-        """Validate settings at startup. Raises SystemExit on invalid config."""
+        """Validate settings at startup. Reports all problems at once.
+
+        Raises SystemExit listing every configuration error found, so the
+        user can fix them in one pass rather than one restart at a time.
+        """
+        errors: list[str] = []
+
         if not self.graph_prefix:
+            errors.append(
+                "GRAPH_PREFIX must not be empty. An empty prefix disables "
+                "the managed-graph safety guard, allowing operations on any "
+                "graph in the account."
+            )
+
+        if errors:
             raise SystemExit(
-                "ERROR: GRAPH_PREFIX must not be empty. "
-                "An empty prefix disables the managed-graph safety guard, "
-                "allowing operations on any graph in the account."
+                "ERROR: invalid configuration:\n"
+                + "\n".join(f"  - {e}" for e in errors)
             )
 
 
--- a/proxy/src/nx_neptune_proxy/routers/graph.py
+++ b/proxy/src/nx_neptune_proxy/routers/graph.py
@@ -40,7 +40,7 @@ def _require_status(graph_id: str, resp: dict) -> str:
 def get_available_actions(graph_id: str):
     """Return valid actions based on the graph's current Neptune status."""
     client = ClientFactory().neptune()
-    resp = client.get_graph(graphIdentifier=graph_id)
+    resp = get_graph_or_exception(client, graph_id)
     # Prefix guard first, before any status handling.
     assert_managed_graph(resp.get("name"))
     status = _require_status(graph_id, resp)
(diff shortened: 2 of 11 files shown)


proxy/src/nx_neptune_proxy/config.py
```diff
@@ -30,3 +30,12 @@ def from_env(cls) -> "Settings":
             region=os.environ.get("AWS_DEFAULT_REGION", os.environ.get("AWS_REGION", "")),
             graph_prefix=os.environ.get("GRAPH_PREFIX", "nxp-"),
         )
+
+    def validate(self) -> None:
+        """Validate settings at startup. Raises SystemExit on invalid config."""
+        if not self.graph_prefix:
+            raise SystemExit(
```
acarbonetto, line 78: I know there's only one item right now, but can we change this to saving a list of configuration errors and we report them all at once?  That way if there are 10 problems with the config, the user doesn't have to fix them one-at-a-time

acarbonetto (approved):

proxy/src/nx_neptune_proxy/utils/aws_helper.py
```diff
@@ -34,7 +34,7 @@ def assert_managed_graph(graph_name: str | None) -> None:
     if not graph_name or not graph_name.startswith(prefix):
         raise HTTPException(
             status_code=403,
-            detail=f"Refusing to delete graph '{graph_name or ''}': not managed by this tool (expected prefix '{prefix}')",
+            detail=f"Refusing to operate on graph '{graph_name or ''}': not managed by this tool (expected prefix '{prefix}')",
```
acarbonetto: ## 403 guard message leaks unmanaged graph names

  The guard returns Refusing to operate on graph '<real-name>': ..., echoing the actual name of graphs the caller doesn't manage. This PR broadens that information disclosure from delete-only to stop/start/actions — three more endpoints. A CSRF-driven tab or any client reaching the proxy can enumerate unmanaged graph names in the account. The name isn't needed for the guard; omit it or return a generic message.

proxy/src/nx_neptune_proxy/routers/graph.py
```diff
@@ -53,9 +54,8 @@ def perform_action(graph_id: str, action: str, background_tasks: BackgroundTasks
         logger.warning(f"Graph {graph_id} returned empty status: {resp}")
         raise HTTPException(status_code=502, detail="Graph returned no status")
 
-    # Prefix guard: only allow destructive actions on graphs managed by this tool
-    if action == "delete":
-        assert_managed_graph(resp.get("name"))
+    # Prefix guard: only allow actions on graphs managed by this tool
+    assert_managed_graph(resp.get("name"))
```
acarbonetto, line 63: In perform_action, the prefix guard runs after the empty-status 502 check; in get_available_actions it runs before the status check. For an unmanaged graph with empty/missing status, perform_action returns 502 'Graph returned no status' before the guard fires, while get_available_actions returns 403 — leaking existence/status differently across endpoints. Move assert_managed_graph above the status check in perform_action so the guard is consistently the first gate.

Otherwise, document reasoning.

proxy/src/nx_neptune_proxy/utils/aws_helper.py
acarbonetto: assert_managed_graph rebuilds a frozen Settings (re-parsing CORS_ALLOWED_ORIGINS, PORT int-cast, region, etc.) on every guarded request just to read graph_prefix. app.py already has a module-level settings — read the prefix from that shared instance instead.

proxy/src/nx_neptune_proxy/config.py
```diff
@@ -30,3 +30,12 @@ def from_env(cls) -> "Settings":
             region=os.environ.get("AWS_DEFAULT_REGION", os.environ.get("AWS_REGION", "")),
             graph_prefix=os.environ.get("GRAPH_PREFIX", "nxp-"),
         )
+
+    def validate(self) -> None:
```
acarbonetto, line 62: Settings.validate() enforces a non-empty prefix against the import-time settings snapshot, but the guard independently re-reads GRAPH_PREFIX via from_env(). They agree today, but the startup validation gives false assurance that the value the guard uses was validated — a future refactor or post-startup mutation could let the guard run on an unvalidated/empty prefix. Validate and enforce the same instance.

proxy/src/nx_neptune_proxy/routers/graph.py
```diff
@@ -29,6 +29,7 @@ def get_available_actions(graph_id: str):
     """Return valid actions based on the graph's current Neptune status."""
     client = ClientFactory().neptune()
     resp = client.get_graph(graphIdentifier=graph_id)
```
acarbonetto: get_available_actions calls client.get_graph directly, while perform_action and the metadata/projection handlers use get_graph_or_exception. A ResourceNotFoundException here only becomes a 404 via the global handler, whereas perform_action handles it explicitly. Route it through get_graph_or_exception for consistency.

proxy/src/nx_neptune_proxy/config.py
andy-k-improving: Fixed — added a validated get_settings() singleton in config.py and migrated all runtime consumers (the guard, plus `app.py/metadata.py/pipeline.py`) to it, so validation and enforcement now share the same instance. 

`from_env()` remains the pure factory. Added tests for same-instance/validate-on-first-access/guard-reads-validated-instance.

proxy/src/nx_neptune_proxy/routers/graph.py
andy-k-improving: Fixed — guard now runs before the status check in perform_action, consistent with get_available_actions, so unmanaged graphs get 403 either way.
Also pulled the duplicated 502 check into a `_require_status helper`.

proxy/src/nx_neptune_proxy/config.py
andy-k-improving: Done — `validate()` now accumulates all errors and reports them together in one bulleted `SystemExit`, so future checks just append to the list instead of failing one-at-a-time.

proxy/src/nx_neptune_proxy/utils/aws_helper.py
andy-k-improving: Done.

proxy/src/nx_neptune_proxy/routers/graph.py
andy-k-improving: Done — `get_available_actions` now fetches via get_graph_or_exception, so a missing graph returns 404 explicitly, consistent with perform_action and the other handlers.

proxy/src/nx_neptune_proxy/utils/aws_helper.py
andy-k-improving: This is now fixed as config object will only be built once during import time.

PR #408, opened 2026-08-17
Feat: Make graph connectivity and deletion protection configurable via env vars

andy-k-improving:
## Summary

Graph creation previously hardcoded `publicConnectivity=true` and `deletionProtection=false`. This makes both configurable via environment variables so users can set them per environment without passing an explicit config dict.

## What changed

Both config builders (`_get_create_instance_config` and `_get_create_instance_with_import_config`) now read two env vars:

| Setting | Default | Env var | Behavior |
|---------|---------|---------|----------|
| `publicConnectivity` | `False` | `NETWORKX_PUBLIC_CONNECTIVITY` | set `true` to enable |
| `deletionProtection` | `True` | `NETWORKX_DELETION_PROTECTION` | set `false` to disable |

Precedence: explicit value in `config` > env var > default. An unrecognized env value falls back to the default, so a typo won't flip a setting unexpectedly.

Defaults changed from the previous values; anyone who wants the old behavior sets the two env vars (the proxy does exactly this — see below).

## Proxy (local dev)

The proxy creates and tears down short-lived graphs from a laptop, so it sets public connectivity on and deletion protection off at launch, rather than changing the library:

- `make dev` / `make run` set both env vars inline
- `make up` sets them and `docker-compose.yml` passes them through from the host
  (unset → library defaults)

## Tests

- Updated the two config-builder tests for the new defaults.
- Added: env override, explicit-beats-env, and unrecognized-value-falls-back.
- 72 tests pass in `test_instance_management.py`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,7 +208,7 @@ environment, or a SageMaker notebook configured for that VPC — with a
 [Neptune Analytics private endpoint](https://docs.aws.amazon.com/neptune-analytics/latest/userguide/vpc.html)
 reachable from the client's subnet and security group.
 
-### Working from a laptop (public connectivity)
+### Connecting to a public graph (local development)
 
 If you are developing locally and connecting directly to the graph from your
 machine, opt into a public endpoint and (optionally) relax deletion protection:
--- a/README.md
+++ b/README.md
@@ -180,6 +180,52 @@ Analytics instance. Make sure your AWS credentials are properly configured and
 your IAM role/user has the required permissions (ReadDataViaQuery,
 WriteDataViaQuery, DeleteDataViaQuery).
 
+## Graph creation defaults
+
+When nx-neptune creates a Neptune Analytics graph for you (via `SessionManager`,
+the instance-management APIs, or the proxy), it applies **secure defaults**:
+
+| Setting | Default | Environment variable | Effect |
+|---|---|---|---|
+| Public connectivity | `false` | `NETWORKX_PUBLIC_CONNECTIVITY` | Graph has **no public endpoint** — reachable only from within its VPC |
+| Deletion protection | `true` | `NETWORKX_DELETION_PROTECTION` | Graph **cannot be deleted** until protection is turned off |
+
+These defaults favor safety over convenience: a freshly created graph is private
+and cannot be accidentally destroyed. Public connectivity is **opt-in** — set
+`NETWORKX_PUBLIC_CONNECTIVITY=true` to enable it. Deletion protection is
+**opt-out** — set `NETWORKX_DELETION_PROTECTION=false` to disable it. Values are
+case-insensitive, and any unrecognized value keeps the secure default.
+
+An explicit value passed in the `config` dict always wins over the environment
+variable, which in turn wins over the default.
+
+### Connecting to a private graph (VPC)
+
+With `NETWORKX_PUBLIC_CONNECTIVITY=false`, the graph is only reachable over a
+private endpoint. The client (your notebook, script, or the proxy) must run
+inside the same VPC as the graph — for example on an EC2 instance, a Cloud9
+environment, or a SageMaker notebook configured for that VPC — with a
+[Neptune Analytics private endpoint](https://docs.aws.amazon.com/neptune-analytics/latest/userguide/vpc.html)
+reachable from the client's subnet and security group.
+
+### Working from a laptop (public connectivity)
+
+If you are developing locally and connecting directly to the graph from your
+machine, opt into a public endpoint and (optionally) relax deletion protection:
+
+```bash
+export NETWORKX_PUBLIC_CONNECTIVITY=true
+export NETWORKX_DELETION_PROTECTION=false
+```
+
+> **Note on the local proxy:** the proxy's `make dev`, `make run`, and `make up`
+> targets already set these two variables, because the proxy is a local
+> development tool that creates and tears down short-lived graphs. If you launch
+> the proxy **outside** the Makefile (e.g. raw `uvicorn`, or a container that
+> does not set them), you get the secure defaults — a private graph you cannot
+> reach from a laptop and that will refuse deletion — so set the variables
+> explicitly in that case.
+
 ## Running tests
 
 Unit tests can be run with make, this runs all tests in the `test` folder:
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -180,6 +180,52 @@ Analytics instance. Make sure your AWS credentials are properly configured and
...
+### Connecting to a private graph (VPC)
+
+With `NETWORKX_PUBLIC_CONNECTIVITY=false`, the graph is only reachable over a
+private endpoint. The client (your notebook, script, or the proxy) must run
+inside the same VPC as the graph — for example on an EC2 instance, a Cloud9
+environment, or a SageMaker notebook configured for that VPC — with a
+[Neptune Analytics private endpoint](https://docs.aws.amazon.com/neptune-analytics/latest/userguide/vpc.html)
+reachable from the client's subnet and security group.
+
+### Working from a laptop (public connectivity)
```
acarbonetto: this is a weird title?

acarbonetto (commented):
I think this is a positive direction for the product, since we want to make it production ready.  We should have a way to test this within an environment that includes a VPC.  Can we look at our CFN and make sure we have a VPC defined (re-using VPC and Subnet if one exists and is provided) and try to connect to the Graph.

nx_neptune/instance_management.py
```diff
@@ -730,9 +730,18 @@ def _get_create_instance_config(graph_name, config=None):
...
+    # Secure by default: private graph + deletion protection. Public connectivity
+    # is opt-in ("true"); deletion protection is opt-out ("false"). Any other
+    # value keeps the secure state. An explicit `config` value wins over both.
+    config.setdefault(
+        "publicConnectivity",
+        os.getenv("NETWORKX_PUBLIC_CONNECTIVITY", "false").strip().lower() == "true",
+    )
     config.setdefault("replicaCount", 0)
-    config.setdefault("deletionProtection", False)
+    config.setdefault(
```
acarbonetto, line 761: New deletionProtection=True default breaks the library's own delete path

  Flipping the create default to deletionProtection=True isn't matched by any change to the delete path. `SessionManager(..., cleanup_task=CleanupTask.DESTROY)` creates a protected graph, then on `__exit__` runs `destroy_all_graphs → _destroy_graphs → delete_na_instance → _delete_na_instance_task`, which calls `client.delete_graph(..., skipSnapshot=True)` with no prior update_graph to clear protection. Neptune Analytics rejects deleting a protected `graph →
  ConflictException`, and the graph is orphaned. Same for the `create_na_instance()/delete_na_instance()` flow in instance_mgmt_lifecycle_demo.ipynb.

  This affects notebooks/session_manager_comprehensive_demo.ipynb and integ_test/session_manager/*. The fix: clear deletion protection via update_graph before delete_graph in the teardown path.

nx_neptune/instance_management.py
```diff
@@ -730,9 +730,18 @@ def _get_create_instance_config(graph_name, config=None):
 
     config = config or {}
     # Ensure mandatory config present.
-    config.setdefault("publicConnectivity", True)
+    # Secure by default: private graph + deletion protection. Public connectivity
+    # is opt-in ("true"); deletion protection is opt-out ("false"). Any other
+    # value keeps the secure state. An explicit `config` value wins over both.
+    config.setdefault(
```
acarbonetto, line 756: New publicConnectivity=False default silently breaks connect-after-create

  A notebook/script that calls create_na_instance() with no config and then connects from a laptop previously got a public endpoint; it now gets a private, VPC-only graph and the connection times out. Only the proxy Makefile sets NETWORKX_PUBLIC_CONNECTIVITY=true — all other existing creation callers regress with no code-level warning.

tests/clients/test_instance_management.py
```diff
@@ -954,9 +954,9 @@ async def test_create_graph_config_base():
     result = _get_create_instance_config("test")
     expected = {
         "graphName": "test",
-        "publicConnectivity": True,
+        "publicConnectivity": False,
```
acarbonetto, line 1021: test_create_graph_config_base / _custom_parameters / test_get_create_instance_with_import_config assert the secure defaults without clearing NETWORKX_* env vars. A developer who follows the README's own laptop instructions (export NETWORKX_PUBLIC_CONNECTIVITY=true, export NETWORKX_DELETION_PROTECTION=false) will see these tests fail spuriously. They should monkeypatch.delenv both vars, like the env-override tests do.

mykola-pereyma (approved):
LGTM — reviewed the diff. Secure-by-default (private graph + deletion protection) with opt-in/opt-out env vars, explicit config taking precedence, and unrecognized values falling back to the secure state. os import is present, and tests cover the env-override, explicit-beats-env, and unrecognized-value cases. Proxy Makefile/compose correctly re-enable public connectivity for local dev.

README.md
andy-k-improving: Fixed

tests/clients/test_instance_management.py
andy-k-improving: Fixed. `test_create_graph_config_base`,  `test_create_graph_config_custom_parameters`, and `test_get_create_instance_with_import_config` now `monkeypatch.delenv` both `NETWORKX_*` vars, 
so they assert the defaults hermetically. Verified they pass even with `NETWORKX_PUBLIC_CONNECTIVITY=true/NETWORKX_DELETION_PROTECTION=false` set in the environment.

nx_neptune/instance_management.py
andy-k-improving: The private-by-default behavior is intended (secure default), with that said, I have added a `logger.info` at create time when publicConnectivity resolves to false, naming the opt-in env var, so the connect-from-outside-VPC timeout is diagnosable. 
Behaviour is unchanged; only the hint is new. Covered by tests (private logs the hint, public stays silent).

nx_neptune/instance_management.py
andy-k-improving: Fixed. `_delete_na_instance_task` now calls `update_graph(deletionProtection=False)` before delete_graph, so the DESTROY/teardown path no longer hits ConflictException on a protected graph.
It's best-effort — if the graph is already unprotected the update is a harmless no-op, and a failure there is logged rather than blocking the delete. Added tests covering the update→delete ordering and the delete-still-proceeds-on-clear-failure case.

PR #410, opened 2026-08-18
Fix: Reuse graphs by exact id and validate graph_name

andy-k-improving:
## Summary

The pipeline previously matched existing graphs by `startswith({prefix}{graph_name})` and reused the first match. With a short or empty `graph_name` this could match more broadly than intended. This makes graph reuse identity-based (exact `graph_id`) and adds input validation on `graph_name`, so reuse is predictable and scoped to the projection's own graph.

## Changes

- **`schemas.py`** — validate `graph_name` (`ProjectionCreate`/`Update`):  min 3 / max 63, pattern `^[A-Za-z0-9][A-Za-z0-9_-]{2,62}$` (upper/lower, digit, `-`, `_`). Empty/short values now return 422.
- **`pipeline.py`** — first run always creates a new graph rather than adopting one by name; retry looks up the recorded `graph_id` exactly and resets only that graph, after confirming it carries the managed prefix. If it's not a managed graph the run stops and the projection is marked `failed` (no reset).  Shared `list_graphs()` behaviour is unchanged.
- **`test_graph_name_collision.py`** (new) — covers the validation rules and the reuse/adoption behaviour.

## Testing

Full proxy suite green: 115 passed (102 existing + 13 new).

### DIFF
--- a/proxy/src/nx_neptune_proxy/routers/schemas.py
+++ b/proxy/src/nx_neptune_proxy/routers/schemas.py
@@ -4,7 +4,19 @@
 from datetime import datetime
 from typing import Optional
 
-from pydantic import BaseModel
+from pydantic import BaseModel, Field
+
+
+# --- graph_name validation ---
+
+# Neptune Analytics graph names in this tool are formed as `{prefix}{graph_name}`.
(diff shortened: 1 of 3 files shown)


mykola-pereyma (approved):
LGTM

PR #411, opened 2026-08-18
Feat: Harden openCypher query construction for algorithm parameters and labels

andy-k-improving:
## Summary

Algorithm parameters and labels were string-interpolated into openCypher without validation or escaping.
This constrains what the library emits into an algorithm CALL to an allowlisted vocabulary, and escapes labels — so generated queries are predictable and well-formed. (Property values and node IDs were already parameterized; this covers the remaining identifiers and algorithm parameters.)

## Allowlist / fail-closed model

Algorithm config is an inline map literal in the CALL clause, so its parts can't be bound as query parameters ($n).
Each is now validated positively — only known values pass, everything else is rejected (fail-closed, ValueError):

- Keys allowlisted against the known parameter names (ALLOWED_ALGO_PARAM_KEYS).
- Enum values (traversalDirection, edge/vertexWeightType) checked against their documented value sets.
- Numeric params must be int/float.
- Identifier params (vertexLabel, writeProperty, edge/vertexWeightProperty) are emitted as escaped double-quoted string literals.
- List values (edgeLabels, sourceNodes, sourceWeights) escape each element.

Node/edge labels and relationship types are backtick-escaped where interpolated.

## Scope

Library-only: opencypher_builder.py + allowlists/enum sets in neptune_constants.py.

## Testing

- Unit (CI): tests/clients/test_opencypher_injection.py, plus updated test_opencypher_builder.py expectations.
- Live graph: all 75 algorithm integ tests pass (label filtering, enum params, write-property mutations all still correct — no regression).

### DIFF
--- a/integ_test/graph_operations/test_security_injection.py
+++ b/integ_test/graph_operations/test_security_injection.py
@@ -139,6 +139,74 @@ class TestInjectionUpdateNode:
         props = results[0]["n"].get("~properties", {})
         assert props["injected"] == payload
 
+class TestInjectionAlgorithmParameters:
+    """Prove label / property-key / algorithm-parameter injection is neutralized.
+
+    These exercise the identifier and algorithm-parameter interpolation paths
+    (vertex_label, edge_labels, write_property) that are NOT covered by the
+    property-value / node-ID parameterization above.
+    """
+
+    def test_pagerank_vertex_label_injection(self, na_graph):
+        import networkx as nx
+
+        # Contains a double-quote (the breakout char for a quoted string literal),
+        # a backslash, and a backtick — so it exercises the double-quote/backslash
+        # escaping applied to identifier-valued algorithm parameters.
+        payload = 'X"`\\ }) MATCH (m) DETACH DELETE m //'
+        # Must never execute the injected DETACH DELETE — either runs harmlessly
+        # (label matches nothing) or raises.
+        try:
+            nx.pagerank(nx.Graph(), backend="neptune", vertex_label=payload)
+        except Exception:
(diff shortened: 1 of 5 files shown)


mykola-pereyma (approved):
LGTM — correct, well-constructed hardening. Escaping order is right, allowlist is tight, fail-closed holds.

A few non-blocking suggestions for follow-up:

**1. Centralize label escaping (drift risk)**
`_escape_labels`/`_escape_identifier` are called individually at ~10 call sites. A future query builder that forgets the call silently reintroduces the vuln. Consider escaping at a single choke point (e.g., always inside `_append_node` and the batch-string helpers) so `escape=False` sites can't drift. Optionally add a lint/test asserting no builder emits a raw label.

**2. Two inconsistent node-ID rendering paths**
`sourceNodes` via `_to_parameter_list` renders as escaped double-quoted literals (`["A","B"]`), while `closeness_centrality_query` renders via `_get_nodes_in_list` as single-quoted regex-validated (`['A','B']`). Both are safe, but the divergence is a maintenance risk. Also verify `sourceWeights` numeric elements stay numeric (not double-quote wrapped).

**3. Error messages echo raw input**
`f"Invalid value {value!r} for parameter {key!r}"` reflects attacker-supplied payloads into exceptions/logs. Consider truncating `{value!r}` for identifier/list values.

**Test suggestions (non-blocking):**
- Add a unicode-escape bypass test (`\u0022`) to lock in the backslash-first escaping order
- Add a trailing-backslash payload test
- Add empty-list (`edgeLabels: []`) case

PR #413, opened 2026-08-18
Feat: Validate identifiers and literals in Athena SQL/DDL generators

andy-k-improving:
## Summary

generate_projection_stmt and generate_create_table_ddl interpolate several caller-supplied values into the SQL/DDL they build; previously only the table name was validated.
This validates the rest so generated statements are well-formed, using an allowlist approach consistent with the existing
_validate_sql_identifier.

## Changes (nx_neptune/utils/utils.py)

- DDL: validate s3_location as an s3:// URI; validate each column name, and
  check each column type against an allowlist (incl. complex types like array<float>, decimal(10,2)).
- Projection: validate col_id/col_label/col_embedding/col_vector_id and each  columns entry as identifiers; validate base_table and join tables (with optional alias); restrict join conditions to equality form; validate  vector_bucket/vector_index/connector_name.
- Shared kinds modelled via a small SqlValueKind enum + one _validate_value helper; invalid input raises ValueError.

## Scope

Library-only. The two import demo notebooks are the only callers and still work
with their existing inputs.

## Testing

- New tests/utils/test_sql_injection.py (23 tests): demo inputs still generate
  SQL/DDL; malformed values are rejected.
- Full suite green (one unrelated pre-existing failure). black/flake8 clean.

### DIFF
--- a/tests/utils/test_sql_injection.py
+++ b/tests/utils/test_sql_injection.py
@@ -1,0 +1,186 @@
+# Copyright 2026 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+"""Tests that generated Athena SQL/DDL validates every caller-supplied
+identifier and literal argument, rejecting values that could break out of the
+surrounding SQL context.
+"""
+
+import pytest
+
+from nx_neptune.utils.utils import (
+    generate_create_table_ddl,
+    generate_projection_stmt,
+)
+
+# The 3-part quoted, aliased table reference used by the S3-vector demo.
+VECTOR_TABLE_REF = '"lambda:my-connector"."my-bucket"."my-index" v'
+
+
+class TestDdlLegitimate:
+    """Legitimate inputs from the import demos still produce DDL."""
+
+    def test_basic_ddl(self):
+        ddl = generate_create_table_ddl(
+            "my_table",
+            "s3://my-bucket/path/",
+            [("id", "string"), ("year", "int"), ("embedding", "array<float>")],
+        )
+        assert "CREATE EXTERNAL TABLE" in ddl
+        assert "`id` string" in ddl
+        assert "LOCATION 's3://my-bucket/path/'" in ddl
+
+    def test_complex_types(self):
+        ddl = generate_create_table_ddl(
+            "t",
+            "s3://my-bucket/p/",
+            [("a", "decimal(10,2)"), ("b", "map<string,int>"), ("c", "array<real>")],
+        )
+        assert "`a` decimal(10,2)" in ddl
+
+
+class TestDdlRejects:
+    def test_s3_location_quote_breakout(self):
+        with pytest.raises(ValueError):
+            generate_create_table_ddl(
+                "t", "s3://b-ucket/p/' ; DROP TABLE x --", [("id", "string")]
+            )
+
+    def test_s3_location_not_s3_uri(self):
+        with pytest.raises(ValueError):
+            generate_create_table_ddl("t", "/local/path", [("id", "string")])
+
+    def test_column_type_statement_injection(self):
+        with pytest.raises(ValueError):
+            generate_create_table_ddl(
+                "t", "s3://my-bucket/p/", [("id", "string); DROP TABLE x --")]
+            )
+
+    def test_column_type_not_allowlisted(self):
+        with pytest.raises(ValueError):
+            generate_create_table_ddl("t", "s3://my-bucket/p/", [("id", "evil")])
+
+    def test_column_name_injection(self):
+        with pytest.raises(ValueError):
+            generate_create_table_ddl(
+                "t", "s3://my-bucket/p/", [("id`; DROP --", "string")]
+            )
+
+    def test_table_name_injection(self):
+        with pytest.raises(ValueError):
+            generate_create_table_ddl(
+                "t; DROP TABLE x --", "s3://my-bucket/p/", [("id", "string")]
+            )
+
+
+class TestProjectionLegitimate:
+    def test_simple_projection(self):
+        stmt = generate_projection_stmt("id", "users", columns=["name", "age"])
+        assert 'id AS "~id"' in stmt
+        assert 'name AS "name"' in stmt
+
+    def test_projection_with_join(self):
+        stmt = generate_projection_stmt(
+            col_id="t.id",
+            col_label="t.masterCategory",
+            col_embedding="v.embedding",
+            columns=["t.gender", "t.subCategory"],
+            base_table="test_embedding_table as t",
+            joins=[(VECTOR_TABLE_REF, "t.id = v.vector_id")],
+        )
+        assert 't.id AS "~id"' in stmt
+        assert "join" in stmt
+        assert "on t.id = v.vector_id" in stmt
+
+    def test_projection_with_udf(self):
+        stmt = generate_projection_stmt(
+            col_id="t.id",
+            col_label="t.masterCategory",
+            col_vector_id="t.id",
+            columns=["t.gender"],
+            base_table="test_embedding_table as t",
+            connector_name="my-connector",
+            vector_bucket="my-bucket",
+            vector_index="my-index",
+        )
+        assert "LAMBDA 'my-connector'" in stmt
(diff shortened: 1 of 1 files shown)


mykola-pereyma (changes requested):
## Security Review: Athena SQL/DDL identifier and literal validation

### Critical: Identifier interpolation without quoting + unverifiable regex

**File:** `nx_neptune/utils/utils.py` (generate_projection_stmt)

Identifiers are interpolated raw (no backtick/quote wrapping): `f'{col_id} AS "~id"'`, `f"{base_table}"`, `{table}`, `on {condition}`. Safety rests entirely on `_SQL_IDENTIFIER_RE` (not visible in this diff). Tests imply the regex accepts quoted dotted colon-bearing segments (e.g., `"lambda:my-connector"."my-bucket"."my-index" v`), which is exactly where quoted-identifier breakout risk lives.

**Recommendation:** Show the regex in the diff (or reference it explicitly), add tests for embedded/doubled quotes in identifiers, and consider wrapping interpolated identifiers as defense-in-depth.

### Critical: `_validate_column_type` is broken for complex types

**File:** `nx_neptune/utils/utils.py` (`_validate_column_type`)

`re.findall(r"[a-zA-Z_]+", value)` extracts ALL alphabetic tokens including struct field **names**. For `struct<name:string>`, it extracts `['struct', 'name', 'string']` — if `name` isn't in the type allowlist, it raises ValueError despite `struct` being allowed. Nested `array<struct<a:int>>` has the same problem.

Additionally, `_COLUMN_TYPE_RE` permits unbalanced `()<>,` and multi-token values like `"int)"` or `"int, string"` — producing malformed DDL.

**Recommendation:** Replace the flat token extraction with a proper type grammar (even a simple recursive regex), or at minimum validate complex types like `struct<field:type>` actually pass your tests (they currently wouldn't).

### Critical: `vector_bucket`/`vector_index` validated conditionally

**File:** `nx_neptune/utils/utils.py` (generate_projection_stmt)

Validation of `vector_bucket` and `vector_index` is nested inside `if col_vector_id:`. If the connector-name code path is reachable without `col_vector_id` set, these values are interpolated unvalidated into `USING EXTERNAL FUNCTION`.

**Recommendation:** Validate at the point of interpolation unconditionally. Add a test for the connector-only path with malicious bucket/index values.

### Suggestions

- `_validate_table_ref` splits on whitespace — fragile for quoted identifiers containing spaces (e.g., `"my table" t` would misparse as 3 parts).
- Four different validation mechanisms (`_validate_value` enum, `_validate_column_type`, `_validate_sql_identifier`, `_validate_table_ref`) — consider a unified dispatch.
- Missing test cases: `struct<field_name:string>`, `array<non_type>`, unbalanced parens `int)`, connector-only-path bucket injection, quoted-space table refs.

### Positives

- `_S3_LOCATION_RE` and `_LITERAL_NAME_RE` correctly exclude quotes/newlines — literal breakout blocked
- `_JOIN_CONDITION_RE` is tight: equality-only, AND-chained, no OR/parens/subqueries
- Fail-closed model consistent across all validators

nx_neptune/utils/utils.py
```diff
@@ -16,12 +16,75 @@
...
+    "string",
+    "varchar",
+    "char",
+    "binary",
+    "date",
+    "timestamp",
+    # complex/parameterized type keywords
+    "array",
+    "map",
+    "struct",
```
noel-improv: Non-blocking, confirming Mykola's point with a repro — `struct` is allowlisted but always rejected, since field names aren't types:

```
generate_create_table_ddl('t','s3://b-x/p/',[('a','struct<name:string,age:int>')])
ValueError: Invalid column type: 'struct<name:string,age:int>' (unknown type 'name').
```

`array<float>`, `map<string,int>`, `decimal(10,2)`, `varchar(10)` all pass. Only `struct` is affected, and nothing calls it today. Drop `"struct"` from the set, or skip field-name positions?

nx_neptune/utils/utils.py
```diff
@@ -16,12 +16,75 @@
...
+class SqlValueKind(Enum):
+    """Kinds of caller-supplied SQL values, each mapped to (regex, error label).
+
+    Used by :func:`_validate_value` to pick the pattern to match and the
+    label to report on failure.
+    """
+
+    JOIN_CONDITION = (_JOIN_CONDITION_RE, "join condition")
+    S3_LOCATION = (_S3_LOCATION_RE, "S3 location")
+    VECTOR_BUCKET = (_LITERAL_NAME_RE, "vector bucket")
```
noel-improv: Nit: `VECTOR_BUCKET`, `VECTOR_INDEX` and `CONNECTOR_NAME` share one regex and differ only in the error label. Would `_validate_literal_name(value, "vector bucket")` do the same job without the enum plus `_validate_value` indirection? `S3_LOCATION` and `JOIN_CONDITION` have their own patterns and would stand alone fine.

nx_neptune/utils/utils.py
```diff
@@ -408,22 +475,28 @@ def generate_projection_stmt(
...
             col_name = col.split(".")[-1].strip('"')
             selects.append(f'{col} AS "{col_name}"')
 
     if col_embedding:
+        _validate_sql_identifier(col_embedding)
         selects.append(f"array_join({col_embedding}, ';') AS \"embedding:vector\"")
 
     if col_vector_id:
+        _validate_sql_identifier(col_vector_id)
+        _validate_value(vector_bucket, SqlValueKind.VECTOR_BUCKET)
```
noel-improv: FWIW this placement looks right to me, contra the review above — `vector_bucket`/`vector_index` are only interpolated in the `col_vector_id` branch. The connector-only path emits `LAMBDA '{connector_name}'` and never touches them, so there's no unvalidated path I can find.

andy-k-improving:
These two helpers (`generate_create_table_ddl` / `generate_projection_stmt`) are only called from two demo notebooks — nothing in the library, proxy, or SessionManager uses them.
Given that footprint, it doesn't make sense to invest in hardening their input validation and maintain that going forward.
I've removed the helpers instead and inlined the SQL directly in the two notebooks, so there's no shared templating layer to validate.
The shared validators used by `_build_sql_statement` are untouched.

Closing this for now and created https://github.com/awslabs/nx-neptune/pull/428 as alternative instead. 


> ## Security Review: Athena SQL/DDL identifier and literal validation
> ### Critical: Identifier interpolation without quoting + unverifiable regex
> **File:** `nx_neptune/utils/utils.py` (generate_projection_stmt)
> 
> Identifiers are interpolated raw (no backtick/quote wrapping): `f'{col_id} AS "~id"'`, `f"{base_table}"`, `{table}`, `on {condition}`. Safety rests entirely on `_SQL_IDENTIFIER_RE` (not visible in this diff). Tests imply the regex accepts quoted dotted colon-bearing segments (e.g., `"lambda:my-connector"."my-bucket"."my-index" v`), which is exactly where quoted-identifier breakout risk lives.
> 
> **Recommendation:** Show the regex in the diff (or reference it explicitly), add tests for embedded/doubled quotes in identifiers, and consider wrapping interpolated identifiers as defense-in-depth.
> 
> ### Critical: `_validate_column_type` is broken for complex types
> **File:** `nx_neptune/utils/utils.py` (`_validate_column_type`)
> 
> `re.findall(r"[a-zA-Z_]+", value)` extracts ALL alphabetic tokens including struct field **names**. For `struct<name:string>`, it extracts `['struct', 'name', 'string']` — if `name` isn't in the type allowlist, it raises ValueError despite `struct` being allowed. Nested `array<struct<a:int>>` has the same problem.
> 
> Additionally, `_COLUMN_TYPE_RE` permits unbalanced `()<>,` and multi-token values like `"int)"` or `"int, string"` — producing malformed DDL.
> 
> **Recommendation:** Replace the flat token extraction with a proper type grammar (even a simple recursive regex), or at minimum validate complex types like `struct<field:type>` actually pass your tests (they currently wouldn't).
> 
> ### Critical: `vector_bucket`/`vector_index` validated conditionally
> **File:** `nx_neptune/utils/utils.py` (generate_projection_stmt)
> 
> Validation of `vector_bucket` and `vector_index` is nested inside `if col_vector_id:`. If the connector-name code path is reachable without `col_vector_id` set, these values are interpolated unvalidated into `USING EXTERNAL FUNCTION`.
> 
> **Recommendation:** Validate at the point of interpolation unconditionally. Add a test for the connector-only path with malicious bucket/index values.
> 
> ### Suggestions
> * `_validate_table_ref` splits on whitespace — fragile for quoted identifiers containing spaces (e.g., `"my table" t` would misparse as 3 parts).
> * Four different validation mechanisms (`_validate_value` enum, `_validate_column_type`, `_validate_sql_identifier`, `_validate_table_ref`) — consider a unified dispatch.
> * Missing test cases: `struct<field_name:string>`, `array<non_type>`, unbalanced parens `int)`, connector-only-path bucket injection, quoted-space table refs.
> 
> ### Positives
> * `_S3_LOCATION_RE` and `_LITERAL_NAME_RE` correctly exclude quotes/newlines — literal breakout blocked
> * `_JOIN_CONDITION_RE` is tight: equality-only, AND-chained, no OR/parens/subqueries
> * Fail-closed model consistent across all validators

PR #414, opened 2026-08-19
Feat: Harden CSV-derived schema handling in export DDL

andy-k-improving:
## Summary

_build_sql_statement derives an Athena CREATE EXTERNAL TABLE schema from the exported CSV header, but only the table name was validated. This validates the header-derived values before they go into the DDL.

## Changes

- instance_management.py: in _build_sql_statement, validate each column name (SQL identifier), each datatype (allowlist), and the s3_location (s3:// URI). Invalid values fail closed with ValueError.
- utils.py: add Neptune export type names that pass through unmapped (bool, byte, short, long) to the type allowlist so normal exports still work.

Library-only, reusing existing validators.

## Testing

- New tests/test_build_sql_statement_injection.py (9 tests): normal export
  header produces expected DDL; malformed names/types rejected.


mykola-pereyma (approved):
LGTM

oussamahansal (approved):
LGTM

nx_neptune/instance_management.py
```diff
@@ -1321,12 +1325,20 @@ def _build_sql_statement(
                 continue
             table_columns[field] = datatype.lower()
 
+    # Column names and datatypes are parsed from the CSV header in the staging
+    # bucket, i.e. from data outside this function's control. Validate each one
+    # before interpolating it into the DDL: names must be safe SQL identifiers
+    # and datatypes must be in the allowlist.
+    for k, v in table_columns.items():
+        _validate_sql_identifier(k)
+        _validate_column_type(v)
```
noel-improv: Blocking: this validates the seed keys too, and both callers seed `~id`/`~from`/`~to`/`~label` (L1220, L1236). `~` isn't a valid identifier char, so every real call now raises:

```
_build_sql_statement(s3, 'my-bucket', '', 'Vertex', [{'Key':'export/Vertex_0.csv'}],
                     {'~id':'string','~label':'string'}, 't_vertices')
ValueError: Invalid SQL identifier: '~id'.
```

Previously these emitted `` `~id` string ``, which Athena accepts. Tests miss it because `_build` seeds `{}`. Could the validation move into the header-parse loop so only header-derived entries are checked?

nx_neptune/instance_management.py
```diff
@@ -1321,12 +1325,20 @@ def _build_sql_statement(
...
     table_schema = ",\n    ".join(f"`{k}` {v}" for k, v in table_columns.items())
```
noel-improv: Blocking-ish: `_COLUMN_TYPE_RE` has no `[` or `]`, and the parser passes `datatype.lower()` straight through, so multi-value export headers fail closed:

```
'~id,tags:String[]'    -> ValueError: Invalid column type: 'string[]'.
'~id,p:String(single)' -> ValueError: Invalid column type: 'string(single)' (unknown type 'single').
```

Do the export CSVs this path reads ever carry `[]` or a cardinality suffix? If yes it needs handling; if no, worth a comment saying so.

nx_neptune/instance_management.py
```diff
@@ -1321,12 +1325,20 @@ def _build_sql_statement(
...
     _validate_sql_identifier(table_name)
     if bucket_folder:
         s3_location = f"s3://{bucket_name}/{bucket_folder}/{prefix}"
     else:
         s3_location = f"s3://{bucket_name}/{prefix}"
+    _validate_s3_location(s3_location)
```
noel-improv: Nit: this lands after the `copy_object`/`delete_object` loop has already moved objects using `prefix`. Moving it above the loop avoids half-migrating the staging bucket on a bad input.

nx_neptune/utils/utils.py
```diff
@@ -22,6 +22,42 @@
...
+    "string",
+    "varchar",
+    "char",
+    "binary",
+    "date",
+    "timestamp",
+}
+
 
 def get_stdout_logger(
```
noel-improv: Rebase note: this branched off `77f6384` rather than #413's head, so it's missing #413's docstring fix (`_validate_sql_value` → `_validate_value`). The `pyproject.toml` / `requirements*.txt` deltas in the diff look like stale-base artifacts too — worth confirming they disappear after the rebase.

PR #416, opened 2026-08-20
Feat: Add per-run access token for proxy API routes

andy-k-improving:
**Summary**

The proxy's `/api/*` routes were reachable by any local process on `localhost:8080`. This adds a per-run bearer token, so the UI and any caller need it to reach those routes.

**Changes**

`auth.py`: random token generated once per process start; `require_token` FastAPI dependency checks `Authorization: Bearer <token>`.
`app.py`: token dependency applied to all four `/api/v0/*` routers; `/health` and `/api/v0/info` stay exempt; token printed to stdout on startup; `index.html` now carries the token via a `<meta>` tag.
`ui-src`: the UI's shared request wrapper reads the token from that tag and attaches it to every API call.

**Scope**

Proxy only. No changes to `nx_neptune` core.

**Testing**

New `tests/test_auth_token.py` (missing/wrong/malformed token → 401, correct token → 200 across all four routers, health/info exempt). Updated existing fixtures to send the token.

### DIFF
--- a/proxy/src/nx_neptune_proxy/app.py
+++ b/proxy/src/nx_neptune_proxy/app.py
@@ -9,12 +9,13 @@ from pathlib import Path
 from urllib.parse import urlparse
 
 from botocore.exceptions import ClientError
-from fastapi import FastAPI, HTTPException, Request
+from fastapi import Depends, FastAPI, HTTPException, Request
 from fastapi.middleware.cors import CORSMiddleware
 from fastapi.responses import FileResponse, JSONResponse
 from fastapi.staticfiles import StaticFiles
 from starlette.middleware.trustedhost import TrustedHostMiddleware
 
+from nx_neptune_proxy.auth import get_token, log_token_notice, require_token
 from nx_neptune_proxy.config import _LOOPBACK_HOSTS, get_settings, normalize_origin
 from nx_neptune_proxy.routers.graph import router as graph_router
 from nx_neptune_proxy.routers.metadata import router as metadata_router
@@ -38,6 +39,17 @@ logging.basicConfig(
     datefmt="%Y-%m-%dT%H:%M:%S",
 )
 logger = logging.getLogger("nx_neptune_proxy")
+
+# --- Proxy access token (per-run bearer token) ---
+# Delivered via the launch URL below, not embedded in any response, so it
+# can't be scraped by reaching the port. In a container stdout hits the log
+# driver, so this URL is log-visible — an accepted tradeoff for a
+# localhost dev tool; the real boundary is the 127.0.0.1 bind + middlewares.
+print(
+    f"\nOpen the proxy UI (token valid for this run only):\n"
+    f"  http://127.0.0.1:{settings.port}/?token={get_token()}\n"
+)
+log_token_notice()
 
 
 # --- App ---
@@ -221,12 +233,12 @@ def info():
     return {"name": "nx-neptune-proxy", "version": "0.1.0"}
 
 
-# --- Routers ---
+# --- Routers (all /api/* routes require the per-run bearer token) ---
 
-app.include_router(metadata_router)
-app.include_router(projection_router)
-app.include_router(project_router)
-app.include_router(graph_router)
+app.include_router(metadata_router, dependencies=[Depends(require_token)])
+app.include_router(projection_router, dependencies=[Depends(require_token)])
+app.include_router(project_router, dependencies=[Depends(require_token)])
+app.include_router(graph_router, dependencies=[Depends(require_token)])
 
 
 # --- Startup: resume stuck deletions ---
@@ -276,6 +288,9 @@ if UI_DIR.exists():
         except ValueError:
             raise HTTPException(status_code=403)
 
+        # Serve a real file if present, else fall back to index.html (SPA
+        # routing). index.html is served as-is — the token is delivered via
+        # the launch URL, not embedded here, so this leaks nothing.
         if file_path.is_file():
             return FileResponse(file_path)
         return FileResponse(ui_root / "index.html")
--- a/proxy/src/nx_neptune_proxy/auth.py
+++ b/proxy/src/nx_neptune_proxy/auth.py
@@ -1,0 +1,59 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Per-run bearer token for the proxy's own HTTP API.
+
+Gates who may call this proxy's /api/* routes. A fresh, random token is
+generated once per process start and kept in memory only — it is never
+written to disk or logged in full.
+"""
+
+import logging
+import secrets
+
+from fastapi import Header, HTTPException
+
+logger = logging.getLogger("nx_neptune_proxy")
+
+# Generated once at import time, i.e. once per process ("per run").
+# Single-worker only: with uvicorn --workers N each worker would generate a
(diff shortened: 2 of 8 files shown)


proxy/src/nx_neptune_proxy/app.py
```diff
@@ -38,6 +39,14 @@
 )
 logger = logging.getLogger("nx_neptune_proxy")
 
+# --- Proxy access token (per-run bearer token) ---
+# Printed directly to stdout (not through the structured logger) so it never
+# ends up in aggregated/forwarded logs. Only a caller with access to this
+# process's own console output, or the bundled UI (which receives it via
+# index.html), can read it.
+print(f"\nProxy access token for this run: {get_token()}\n")
```
acarbonetto: #### The token is readable without the token

The SPA catch-all (spa_fallback) has no require_token dependency, and it serves index.html with the token injected into a <meta> tag. So any client that can reach the port does GET / (no auth header), gets 200, scrapes the token from the body, and then calls any /api/* route. The docstring's claim that only "the console output or the bundled UI" can read the token is false — any non-browser local caller can.

  This is worth pausing on, because it reframes the whole mechanism. The Same-Origin Policy only protects the token against a browser on another origin reading the response body. It does not protect against a non-browser client (curl, another local process, a rebinding attacker's server-side fetch) that simply reads the raw HTML. So the token defends against cross-origin browser tabs but not against any local process — which matches what I flagged earlier: this is CSRF/rebinding hardening, not an access boundary. The reviewer here has demonstrated the concrete bypass.

acarbonetto: Token printed to stdout leaks into container logs

Containerized (Dockerfile present, /app/proxy/ui fallback) means stdout → docker logs/journald/CloudWatch, contradicting the "never ends up in aggregated logs" comment. New wrinkle: because the print runs at import time, it also prints the token during CI test collection whenever nx_neptune_proxy.app is imported. My proposed file-based fix (path-only to stdout, secret to a 0600 file) addresses both.

mykola-pereyma (changes requested):
## Security Review: Per-run access token

### Critical: Token delivery defeats the access control

The token is embedded in `index.html` served by the **unauthenticated** SPA catch-all (`/{path:path}`) on the same port as the protected `/api/*` routes. Any client reaching the port can:

```
curl http://host:8080/ | grep nx-neptune-proxy-token
```

and extract the token to call every protected route. There is no scenario where an attacker can reach `/api/*` but not `/`.

The docstring claims the token is "only readable by same-origin JavaScript (browser SOP)" — this is incorrect for non-browser clients. SOP only constrains browser JS cross-origin reads; `curl`/scripts are not subject to SOP.

**Recommendation:** Either:
1. Don't embed the token in an unauthenticated response (e.g., operator copies from stdout, or use an HttpOnly SameSite=Strict cookie)
2. Accept this is NOT a network auth boundary and document that protection comes from localhost binding + CSRF/CORS only

If option 2: remove the SOP claim from the docstring and update the threat model to state this clearly.

### Additional findings:

- **Log exposure:** `print(token)` to stdout is captured by Docker's log driver and forwarded to CloudWatch/aggregation — the comment claiming it "never ends up in aggregated logs" is false in the container deployment.
- **Multi-worker fragility:** `_TOKEN` is a per-process global. Under `uvicorn --workers N`, each worker generates a different token; the UI served by worker A gets 401 on worker B.
- **Non-ASCII 500:** `compare_digest` raises `TypeError` on non-ASCII bearer values, producing 500 instead of 401. Wrap in try/except or compare on `.encode()` bytes.
- **`/docs` unauthenticated:** OpenAPI schema exposed without token. Low risk but should be intentional.
- **Threat model not updated** to reflect what this token does and does not protect against.

### Positives:
- Token generation (`secrets.token_urlsafe(32)`) and comparison (`compare_digest`) are cryptographically sound
- Test coverage is thorough (missing/wrong/malformed/correct/exempt all tested)
- UI correctly attaches Bearer header from the meta tag

andy-k-improving:
**Critical — token delivery via unauthenticated catch-all**

Went with **option 1**: the token is no longer embedded in any response. 
`index.html` is now served as-is (via `FileResponse`, no meta-tag injection), and the token is delivered out-of-band through a launch URL printed to stdout — the UI reads it from the URL's query string once, strips it from the address bar (`history.replaceState`), and keeps it in memory only (no cookie, no local/sessionStorage). 
`curl http://host:8080/` no longer yields the token. The false SOP claim in the docstring is removed.

**Log exposure**

Accepted and documented rather than engineered around.
This is a localhost-bound dev tool; in a container the launch URL on stdout is log-visible, so the comment now states that explicitly and notes the real boundary is the `127.0.0.1` bind + the CORS/Origin/Host middlewares, not the token. 
Removed the false "never ends up in aggregated logs" claim.

**Non-ASCII 500**

Fixed — `require_token` now compares on `.encode()` bytes, so a non-ASCII bearer value fails cleanly with 401 instead of raising `TypeError`.
Added a test that calls the dependency directly (HTTP clients won't transport non-ASCII header bytes, so this exercises the actual code path).

**Multi-worker fragility**

Documented for now — the token is per-process and the shipped launch paths run a single worker. A follow-up PR makes `python -m studio_main` the sole entry point (Dockerfile + Makefile both repointed) and pins a single worker, removing the `--workers` path entirely.

**/docs unauthenticated**

Already disabled 



> ## Security Review: Per-run access token
> ### Critical: Token delivery defeats the access control
> The token is embedded in `index.html` served by the **unauthenticated** SPA catch-all (`/{path:path}`) on the same port as the protected `/api/*` routes. Any client reaching the port can:
> 
> ```
> curl http://host:8080/ | grep nx-neptune-proxy-token
> ```
> 
> and extract the token to call every protected route. There is no scenario where an attacker can reach `/api/*` but not `/`.
> 
> The docstring claims the token is "only readable by same-origin JavaScript (browser SOP)" — this is incorrect for non-browser clients. SOP only constrains browser JS cross-origin reads; `curl`/scripts are not subject to SOP.
> 
> **Recommendation:** Either:
> 
> 1. Don't embed the token in an unauthenticated response (e.g., operator copies from stdout, or use an HttpOnly SameSite=Strict cookie)
> 2. Accept this is NOT a network auth boundary and document that protection comes from localhost binding + CSRF/CORS only
> 
> If option 2: remove the SOP claim from the docstring and update the threat model to state this clearly.
> 
> ### Additional findings:
> * **Log exposure:** `print(token)` to stdout is captured by Docker's log driver and forwarded to CloudWatch/aggregation — the comment claiming it "never ends up in aggregated logs" is false in the container deployment.
> * **Multi-worker fragility:** `_TOKEN` is a per-process global. Under `uvicorn --workers N`, each worker generates a different token; the UI served by worker A gets 401 on worker B.
> * **Non-ASCII 500:** `compare_digest` raises `TypeError` on non-ASCII bearer values, producing 500 instead of 401. Wrap in try/except or compare on `.encode()` bytes.
> * **`/docs` unauthenticated:** OpenAPI schema exposed without token. Low risk but should be intentional.
> * **Threat model not updated** to reflect what this token does and does not protect against.
> 
> ### Positives:
> * Token generation (`secrets.token_urlsafe(32)`) and comparison (`compare_digest`) are cryptographically sound
> * Test coverage is thorough (missing/wrong/malformed/correct/exempt all tested)
> * UI correctly attaches Bearer header from the meta tag

mykola-pereyma (approved):
LGTM

oussamahansal (approved):
LGTM

PR #420, opened 2026-08-21
Feat: Add TrustedHostMiddleware to block DNS rebinding

andy-k-improving:
## Summary

Adds two complementary, defense-in-depth middlewares to the proxy, both driven from a single origins list:

- **`TrustedHostMiddleware`** validates the `Host` header against an allowlist to block **DNS rebinding**. The `Host` header alone doesn't confirm a request was actually meant for this server — rebinding can trick a client into sending a request that lands here while claiming a hostname it doesn't recognize.
- **`origin_validation`** validates the `Origin` header **server-side** against the CORS allowlist, rejecting disallowed origins with a 403 before the handler runs. This enforces the allowlist on the server rather than relying on the browser (CORS response headers are only browser-enforced and don't stop a request from executing).

## Changes

- **`config.py`**: `allowed_origins` and `trusted_hosts` are now derived once, together, from the single `CORS_ALLOWED_ORIGINS` list — so the two allowlists can't drift. 
`allowed_origins` holds the full normalized origins (used by CORS and the Origin check); `trusted_hosts` holds those origins' hostnames plus the loopback set (used by the Host check).
Loopback (`127.0.0.1`, `localhost`) is always trusted, which covers the normal local/Docker case.

- **`app.py`**: registers `TrustedHostMiddleware` (Host check) and the `origin_validation` middleware (server-side Origin check). Host matching is case-insensitive.
IPv6 is excluded from the Host allowlist for now — Starlette can't parse IPv6 `Host` headers (starlette#3471) — and fails closed: all IPv6 hosts are denied, not allowed, until a fixed release is available.

- **Tests**: fixtures updated to address the proxy with a recognized hostname; coverage for allowed/rejected hosts, the IPv6 exclusion, and server-side Origin validation.

## Testing

113 passed.

### DIFF
--- a/proxy/README.md
+++ b/proxy/README.md
@@ -56,6 +56,6 @@ cd ui-src && npm run dev
 | Environment Variable | Default | Description |
 |---|---|---|
 | `PORT` | `8080` | Server port |
-| `CORS_ALLOWED_ORIGINS` | `http://localhost:5173` | Allowed CORS origins (dev) |
+| `CORS_ALLOWED_ORIGINS` | *(none)* | Comma-separated list of allowed browser origins (full `scheme://host[:port]`, e.g. `https://app.example.com`). This is the single source of truth for origins: it drives the CORS response headers, the server-side Origin check, and the hostnames trusted in the `Host` header. Loopback origins are always allowed, so no configuration is needed for local/Docker use. |
 | `LOG_LEVEL` | `info` | Logging level |
 | `NX_NEPTUNE_DB_PATH` | `~/.nx-neptune/proxy.db` | SQLite database path |
--- a/proxy/docker-compose.yml
+++ b/proxy/docker-compose.yml
@@ -10,6 +10,9 @@ services:
       - AWS_SECRET_ACCESS_KEY=${AWS_SECRET_ACCESS_KEY}
       - AWS_SESSION_TOKEN=${AWS_SESSION_TOKEN}
       - AWS_DEFAULT_REGION=${AWS_DEFAULT_REGION:-us-west-2}
+      # Single source of truth for browser origins: drives CORS, the
+      # server-side Origin check, and the trusted Host hostnames. Loopback
+      # is always allowed; these entries cover the Graph Explorer sidecar.
       - CORS_ALLOWED_ORIGINS=http://localhost:8080,https://localhost:6373
       - LOG_LEVEL=info
       # Pass-through from the host env. If unset, the library applies its secure
--- a/proxy/src/nx_neptune_proxy/app.py
+++ b/proxy/src/nx_neptune_proxy/app.py
@@ -46,7 +46,7 @@ app = FastAPI(title="nx-neptune-proxy", version="0.1.0", docs_url=None, redoc_ur
 
 # --- TrustedHost middleware (blocks DNS rebinding) ---
 #
-# TODO: IPv6 is not supported until encode/starlette#3357 is fixed
+# TODO: IPv6 is not supported until starlette#3471 is fixed
 # (Starlette can't parse IPv6 Host headers, so any IPv6 entry here is
 # dead code). Re-add "::1" once a fixed release is available. No security
 # leak: all IPv6 hosts are denied in the meantime, not allowed.
--- a/proxy/src/nx_neptune_proxy/app.py
+++ b/proxy/src/nx_neptune_proxy/app.py
@@ -46,15 +46,16 @@ app = FastAPI(title="nx-neptune-proxy", version="0.1.0", docs_url=None, redoc_ur
 
 # --- TrustedHost middleware (blocks DNS rebinding) ---
 #
-# Known limitation: Starlette's TrustedHostMiddleware parses the Host header
-# via `.split(":")[0]`, which returns "[" for any IPv6 literal (e.g.
-# "[::1]:8080"), never a value that can match an allowlist entry. This means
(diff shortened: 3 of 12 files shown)


mykola-pereyma (changes requested):
## Security Review: TrustedHostMiddleware + origin_validation

### Critical: Undisclosed `origin_validation` middleware conflicts with CORS allowlist

**File:** `proxy/src/nx_neptune_proxy/app.py` (origin_validation middleware registration)

`origin_validation` checks the Origin header against `trusted_hosts`, which is **independent** of `CORS_ALLOWED_ORIGINS`. A legitimately CORS-allowed origin will be rejected 403 by `origin_validation` unless ALSO added to `TRUSTED_HOSTS`. Two allowlists must stay in sync with no cross-check and no error message explaining the conflict.

**Fix:** Either derive origin_validation hosts from the CORS allowlist, merge the two concepts, or add a startup warning when they diverge.

### Critical: Undisclosed scope

PR titled "Add TrustedHostMiddleware" but introduces a whole second security middleware (`origin_validation` + `test_origin_validation.py`) with different semantics (hostname-only matching, drops scheme/port). This should be described in the PR or split.

### Suggestions

**`config.py` (trusted_hosts construction):**
Case-sensitivity inconsistency: `TrustedHostMiddleware` is case-sensitive and `TRUSTED_HOSTS` values aren't lowercased, while `origin_validation` uses `urlparse` which lowercases. Normalize to lowercase at parse time.

**Tests:**
Missing edge cases: uppercase Host header, empty/missing Host, whitespace-only Host, Host with suffix that could confuse substring matching. Only `/health` is tested against the middleware — no proof it covers the SPA catch-all route.

**Middleware ordering (`app.py`):**
`add_middleware` inserts at front, so execution order is: csrf → origin_validation → CORS → TrustedHost (innermost). CORS preflight `OPTIONS` short-circuits before TrustedHost runs. Functionally safe but likely unintentional — consider an ordering comment.

### Positives
- Core TrustedHostMiddleware is correct and blocks DNS rebinding
- Default `{127.0.0.1, localhost}` is right for loopback proxy  
- IPv6 exclusion fails CLOSED (safe default, documented with Starlette #3357 ref)
- Tests cover allowed/rejected hosts and IPv6 exclusion

andy-k-improving:
**Undisclosed origin_validation middleware conflicts with CORS allowlist:**
Derived `allowed_origins` and `trusted_hosts` once from the single `CORS_ALLOWED_ORIGINS` list in `from_env`, so there's no separate list to sync and they can't diverge.

**Undisclosed scope:**
Kept both middlewares but updated the description to disclose them — they check different headers (`Host` vs `Origin`) and now share the same origins-derived config, so I'd rather not split.

**Case-sensitivity inconsistency:**
Config values are normalized, but Starlette compares the `Host` header case-sensitively, so I'm accepting case-sensitive Host matching for now and added tests documenting it; `origin_validation` stays case-insensitive.

**Missing test edge cases:**
Added tests for uppercase/empty/whitespace/prefix/suffix-confusion Hosts (all rejected) plus a class proving the Host check covers the SPA catch-all route, not just `/health`.

**Middleware ordering:**
Added an ordering comment in `app.py`; also verified a bad-Host `OPTIONS` still returns 400, so the preflight isn't short-circuited past the Host check.

> ## Security Review: TrustedHostMiddleware + origin_validation
> ### Critical: Undisclosed `origin_validation` middleware conflicts with CORS allowlist
> **File:** `proxy/src/nx_neptune_proxy/app.py` (origin_validation middleware registration)
> 
> `origin_validation` checks the Origin header against `trusted_hosts`, which is **independent** of `CORS_ALLOWED_ORIGINS`. A legitimately CORS-allowed origin will be rejected 403 by `origin_validation` unless ALSO added to `TRUSTED_HOSTS`. Two allowlists must stay in sync with no cross-check and no error message explaining the conflict.
> 
> **Fix:** Either derive origin_validation hosts from the CORS allowlist, merge the two concepts, or add a startup warning when they diverge.
> 
> ### Critical: Undisclosed scope
> PR titled "Add TrustedHostMiddleware" but introduces a whole second security middleware (`origin_validation` + `test_origin_validation.py`) with different semantics (hostname-only matching, drops scheme/port). This should be described in the PR or split.
> 
> ### Suggestions
> **`config.py` (trusted_hosts construction):** Case-sensitivity inconsistency: `TrustedHostMiddleware` is case-sensitive and `TRUSTED_HOSTS` values aren't lowercased, while `origin_validation` uses `urlparse` which lowercases. Normalize to lowercase at parse time.
> 
> **Tests:** Missing edge cases: uppercase Host header, empty/missing Host, whitespace-only Host, Host with suffix that could confuse substring matching. Only `/health` is tested against the middleware — no proof it covers the SPA catch-all route.
> 
> **Middleware ordering (`app.py`):** `add_middleware` inserts at front, so execution order is: csrf → origin_validation → CORS → TrustedHost (innermost). CORS preflight `OPTIONS` short-circuits before TrustedHost runs. Functionally safe but likely unintentional — consider an ordering comment.
> 
> ### Positives
> * Core TrustedHostMiddleware is correct and blocks DNS rebinding
> * Default `{127.0.0.1, localhost}` is right for loopback proxy
> * IPv6 exclusion fails CLOSED (safe default, documented with Starlette #3357 ref)
> * Tests cover allowed/rejected hosts and IPv6 exclusion

mykola-pereyma (approved):
LGTM

PR #421, opened 2026-08-21
Feat: Introduce a startup check for the proxy bind address

andy-k-improving:
**Summary**

This introduces a startup check that runs before the proxy begins serving. The bind address is now a configurable setting that defaults to loopback, and the check confirms it before startup, asking for an explicit opt-in before binding to a non-loopback address. Local runs work out of the box, and binding elsewhere is always an intentional choice.

**Changes**

`__main__.py`: new startup entrypoint that runs the check and then starts the server. The proxy is now started through this entrypoint.
`config.py`: the bind address is now a setting (`HOST`, defaulting to loopback), with a companion opt-in (`ALLOW_NON_LOOPBACK_BIND`) for binding elsewhere.
`Dockerfile`: starts the proxy through the new entrypoint and sets the address and opt-in explicitly, since a container binds to all interfaces for its published port to be reachable, with a note explaining why.
Tests covering the startup check, the entrypoint behavior, and the new settings.

**Scope**

Proxy only. The development workflow (`make dev`) is unchanged.

**Testing**

* Make dev and docker-compose continue to work without issue.

### DIFF
--- a/proxy/Dockerfile
+++ b/proxy/Dockerfile
@@ -7,6 +7,12 @@ COPY proxy/ ./proxy/
 RUN pip install --no-cache-dir ./nx_neptune_src ./proxy && useradd --create-home appuser
 USER appuser
 EXPOSE 8080
+# The container must bind 0.0.0.0 so Docker's published-port forwarding can
+# reach it (container-internal 127.0.0.1 is unreachable from the host). Real
+# network exposure is controlled by how the port is published (see the run
+# target's -p 127.0.0.1:8080:8080). ALLOW_NON_LOOPBACK_BIND acknowledges this
+# deliberate bind so the startup guard permits it.
+ENV HOST=0.0.0.0 ALLOW_NON_LOOPBACK_BIND=1
 HEALTHCHECK --interval=10s --timeout=3s --retries=3 \
     CMD wget -qO /dev/null http://localhost:8080/health
-CMD ["uvicorn", "nx_neptune_proxy.app:app", "--host", "0.0.0.0", "--port", "8080"]
+CMD ["python", "-m", "nx_neptune_proxy"]
--- a/proxy/src/nx_neptune_proxy/__main__.py
+++ b/proxy/src/nx_neptune_proxy/__main__.py
@@ -1,0 +1,55 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Process entrypoint: validate the bind address, then launch uvicorn.
+
+Run as ``python -m nx_neptune_proxy``. This wraps the uvicorn launch so the
+bind address is validated *before* the socket is opened. Because the same
+value is both checked here and passed to ``uvicorn.run``, the address that
+gets validated is guaranteed to be the address that gets bound.
+
+"""
+
+import sys
+
+import uvicorn
+
+from nx_neptune_proxy.config import Settings
+
+# A request addressed to any of these never leaves the local machine.
+_LOOPBACK_HOSTS = frozenset({"127.0.0.1", "::1", "localhost"})
+
+
+def check_bind_allowed(host: str, allow_non_loopback: bool) -> str | None:
+    """Return an error message if binding to ``host`` is not allowed, else None.
+
+    Binding to loopback is always allowed. Binding to any other address is
+    allowed only when explicitly acknowledged via ``allow_non_loopback``.
+    """
+    if host in _LOOPBACK_HOSTS:
+        return None
+    if allow_non_loopback:
+        return None
+    return (
+        f"Refusing to start: HOST={host!r} is not a loopback address and "
+        "ALLOW_NON_LOOPBACK_BIND is not set. This would expose the proxy on "
+        "the network. Bind to 127.0.0.1, or set ALLOW_NON_LOOPBACK_BIND=1 to "
+        "expose it deliberately (e.g. inside a container whose published "
+        "ports are restricted)."
+    )
+
+
+def main() -> None:
+    settings = Settings.from_env()
+    error = check_bind_allowed(settings.host, settings.allow_non_loopback_bind)
+    if error:
+        sys.exit(error)
+    uvicorn.run(
+        "nx_neptune_proxy.app:app",
+        host=settings.host,
+        port=settings.port,
+    )
+
+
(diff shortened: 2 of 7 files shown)


mykola-pereyma (changes requested):
## Security Review: Proxy bind address startup check

### Critical: Bearer token served unauthenticated — access control is illusory

**File:** `proxy/src/nx_neptune_proxy/app.py` (`_serve_index` catch-all `/{path:path}`)

The per-run bearer token is embedded in `index.html` via a `<meta>` tag, served by the **unauthenticated** SPA catch-all on the same port as the protected API. Any client reaching the port can:

```
curl http://host:8080/ | grep nx-neptune-proxy-token
```

and call every `/api/*` route. There is no scenario where an attacker can reach `/api/*` but not `/`. The auth.py docstring claims the token is "only readable by same-origin JavaScript (browser SOP)" — this is incorrect for non-browser clients.

**Impact:** The proxy runs arbitrary user Athena SQL and creates/deletes Neptune graphs with the host's AWS credentials.

**Fix:** Don't embed the token in an unauthenticated response. Options:
1. Operator copies token from stdout into the UI (Jupyter-style)
2. HttpOnly SameSite=Strict cookie set only via same-origin navigation
3. Accept this is NOT network auth and document that protection is localhost binding + CSRF/CORS only — remove the false SOP claim

### Critical: Bind guard is bypassable and Docker disables it

**File:** `proxy/src/nx_neptune_proxy/__main__.py` (bind check)
**File:** `proxy/Dockerfile` (`ENV HOST=0.0.0.0 ALLOW_NON_LOOPBACK_BIND=1`)

The guard only runs via `python -m nx_neptune_proxy`; launching uvicorn directly (the `dev` Make target) bypasses it. The Dockerfile permanently disables it with `HOST=0.0.0.0` + `ALLOW_NON_LOOPBACK_BIND=1`. A broad `docker -p 8080:8080` exposes an unauthenticated credential-bearing proxy to the network (combines with finding above).

**Fix:** Enforce at middleware level (reject non-loopback source IPs unless opted in), or prominently document that non-loopback binding exposes the token.

### Suggestions

- **Log exposure:** `print(token)` to stdout is captured by Docker's log driver → CloudWatch. The comment claiming it "never ends up in aggregated logs" is false in container deployments.
- **Error sanitization incomplete:** `proxy/src/nx_neptune_proxy/utils/sanitize.py` — bare 12-digit account IDs, bucket names, and role/session names outside ARN form pass through.
- **SQLite concurrency:** `check_same_thread=False` with connection-per-call may hit "database is locked" under load.
- **Deprecated `on_event("startup")`:** Use lifespan context manager (FastAPI 0.95+ recommendation).

### Positives
- SQL injection properly mitigated (parameterized queries + column allowlist)
- Path traversal has two layers (reject '..' + resolve/relative_to)
- No frontend XSS (React escaping, no dangerouslySetInnerHTML)
- Dockerfile: non-root appuser, slim image, healthcheck, no baked secrets
- Auth primitives correct: `secrets.token_urlsafe(32)` + `compare_digest`
- Comprehensive security test suite

andy-k-improving:
Updated to have 2 tiers checks with the first being the startup check and second on the middleware level to examine the incoming request source IP. 

Bearer token and Log exposure being addressed under https://github.com/awslabs/nx-neptune/pull/416

> ## Security Review: Proxy bind address startup check
> ### Critical: Bearer token served unauthenticated — access control is illusory
> **File:** `proxy/src/nx_neptune_proxy/app.py` (`_serve_index` catch-all `/{path:path}`)
> 
> The per-run bearer token is embedded in `index.html` via a `<meta>` tag, served by the **unauthenticated** SPA catch-all on the same port as the protected API. Any client reaching the port can:
> 
> ```
> curl http://host:8080/ | grep nx-neptune-proxy-token
> ```
> 
> and call every `/api/*` route. There is no scenario where an attacker can reach `/api/*` but not `/`. The auth.py docstring claims the token is "only readable by same-origin JavaScript (browser SOP)" — this is incorrect for non-browser clients.
> 
> **Impact:** The proxy runs arbitrary user Athena SQL and creates/deletes Neptune graphs with the host's AWS credentials.
> 
> **Fix:** Don't embed the token in an unauthenticated response. Options:
> 
> 1. Operator copies token from stdout into the UI (Jupyter-style)
> 2. HttpOnly SameSite=Strict cookie set only via same-origin navigation
> 3. Accept this is NOT network auth and document that protection is localhost binding + CSRF/CORS only — remove the false SOP claim
> 
> ### Critical: Bind guard is bypassable and Docker disables it
> **File:** `proxy/src/nx_neptune_proxy/__main__.py` (bind check) **File:** `proxy/Dockerfile` (`ENV HOST=0.0.0.0 ALLOW_NON_LOOPBACK_BIND=1`)
> 
> The guard only runs via `python -m nx_neptune_proxy`; launching uvicorn directly (the `dev` Make target) bypasses it. The Dockerfile permanently disables it with `HOST=0.0.0.0` + `ALLOW_NON_LOOPBACK_BIND=1`. A broad `docker -p 8080:8080` exposes an unauthenticated credential-bearing proxy to the network (combines with finding above).
> 
> **Fix:** Enforce at middleware level (reject non-loopback source IPs unless opted in), or prominently document that non-loopback binding exposes the token.
> 
> ### Suggestions
> * **Log exposure:** `print(token)` to stdout is captured by Docker's log driver → CloudWatch. The comment claiming it "never ends up in aggregated logs" is false in container deployments.
> * **Error sanitization incomplete:** `proxy/src/nx_neptune_proxy/utils/sanitize.py` — bare 12-digit account IDs, bucket names, and role/session names outside ARN form pass through.
> * **SQLite concurrency:** `check_same_thread=False` with connection-per-call may hit "database is locked" under load.
> * **Deprecated `on_event("startup")`:** Use lifespan context manager (FastAPI 0.95+ recommendation).
> 
> ### Positives
> * SQL injection properly mitigated (parameterized queries + column allowlist)
> * Path traversal has two layers (reject '..' + resolve/relative_to)
> * No frontend XSS (React escaping, no dangerouslySetInnerHTML)
> * Dockerfile: non-root appuser, slim image, healthcheck, no baked secrets
> * Auth primitives correct: `secrets.token_urlsafe(32)` + `compare_digest`
> * Comprehensive security test suite

mykola-pereyma (approved):
LGTM

oussamahansal (approved):
LGTM

PR #428, opened 2026-08-25
Misc: Remove Athena SQL/DDL generator helpers from the library

andy-k-improving:
### Summary

Removes `generate_create_table_ddl` and `generate_projection_stmt` from `nx_neptune/utils/utils.py`. These SQL/DDL templating helpers had only two callers — both demo notebooks — which is too narrow to justify maintaining them as library API. The notebooks now contain the SQL inline.

### Changes

- **`nx_neptune/utils/utils.py`** — removed the two generator functions.
- **`notebooks/import_open_search_embedding_demo.ipynb`, `notebooks/import_s3_vector_embedding_demo.ipynb`** — replaced the generator calls with equivalent inline SQL; dropped the now-redundant `columns` list and imports. Generated SQL is unchanged.
- **`tests/utils/test_sql_injection.py`** — removed; it covered only the deleted helpers.

### Testing

- `test_build_sql_statement_injection.py` + `test_utils.py` pass (45 tests), confirming the retained validators are intact.
- Inline notebook SQL verified byte-identical to the generator output.


mykola-pereyma (approved):
LGTM — clean removal that resolves the concerns raised on the earlier SQL/DDL-generator hardening (#413) by deletion rather than repair. The generators had only two callers (both demo notebooks), which is too narrow to maintain as library API. Verified the inline notebook SQL matches the removed template structure, and the retained validators are intact.

Confirmed against the branch source:
- The three retained validators (`_validate_sql_identifier`, `_validate_column_type`, `_validate_s3_location`) are all actively used by `_build_sql_statement` (the CSV-header-derived path, which is the actually-untrusted input) — no dead code introduced by this removal.
- `_validate_column_type` here uses a clean exact-match scalar allowlist (`_ALLOWED_COLUMN_TYPES`), which correctly rejects both malformed values and complex types for the scalar-only CSV-export path.

Two follow-up notes (non-blocking):

1. **Close #413 as superseded.** It hardened `generate_projection_stmt`/`generate_create_table_ddl`, which this PR removes entirely. Both target `ft-hf-graph-studio-p1` and conflict — once this merges, #413's target no longer exists.

2. **Mind the merge order vs #413.** This PR removes `tests/utils/test_sql_injection.py`; #413 adds it. If #413 were to merge after this, it would re-introduce tests for deleted functions.

PR #429, opened 2026-08-25
Misc: Add proxy-fmt / proxy-lint make targets and CI lint job

andy-k-improving:
### Summary :memo:

Adds format/lint tooling for the `proxy/` module, which had no lint coverage before.

- `Makefile`: `proxy-fmt` (isort + black) and `proxy-lint` (flake8 + black --check + mypy) over `proxy/src/ proxy/tests/`, mirroring the root `fmt`/`lint`.

- `.github/workflows/main.yml`: new `proxy_lint` job running `make install-dev` then `make proxy-lint`.

PR #430, opened 2026-08-26
Misc: Make the proxy module pass `make proxy-lint`

andy-k-improving:
### Summary :memo:

Follow-up to #429 — makes the proxy pass the lint gate (was red on pre-existing debt).

- Formatting (`make proxy-fmt`): black + isort across the module; mechanical, no behavior change.

- mypy fixes, minimal/behavior-preserving: `projection_store.py` 7 params `str = None` → `Optional[str] = None`; 4 `# type: ignore` at use sites; `all_results: list` annotation.

- `opencypher_builder.py`: widened `_escape_labels` return to `list | str` (pre-existing mypy error found via `make test`).

Testing: `make proxy-lint` passes (flake8 + black + mypy); proxy suite 137 passed; root `make test` 434 passed.

### DIFF
--- a/nx_neptune/clients/opencypher_builder.py
+++ b/nx_neptune/clients/opencypher_builder.py
@@ -87,7 +87,7 @@ def _truncate_for_error(value: Any, limit: int = 40) -> str:
     return text
 
 
-def _escape_labels(labels) -> list:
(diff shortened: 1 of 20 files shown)


andy-k-improving:
Merging this as CI can only be test on P1 branch.

PR #431, opened 2026-08-26
Misc: Apply proxy-fmt (black + isort) to proxy module (P2)

andy-k-improving:
## Summary

Runs `make proxy-fmt` (black + isort) across the proxy module to bring it into line with the repo's formatting rules. **Formatting only — no logic or behavior changes.**

## Changes

Reformatted files (long-line wrapping, import ordering):

- `proxy/src/nx_neptune_proxy/routers/project_io.py`
- `proxy/src/nx_neptune_proxy/routers/schemas.py`
- `proxy/src/nx_neptune_proxy/utils/aws_helper.py`
- `proxy/src/nx_neptune_proxy/utils/sanitize.py`
- `proxy/tests/test_column_allowlist.py`
- `proxy/tests/test_no_secrets.py`
- `proxy/tests/test_project_io.py`
- `proxy/tests/test_project_io_s3.py`
- `proxy/tests/test_projection.py`
- `proxy/tests/test_sql_injection.py`

## Testing

- `make proxy-lint` passes (flake8 + black --check + mypy)

### DIFF
--- a/proxy/src/nx_neptune_proxy/routers/project_io.py
+++ b/proxy/src/nx_neptune_proxy/routers/project_io.py
@@ -9,23 +9,30 @@ and re-importing it to recreate the project in another environment.
 
 from __future__ import annotations
 
-from typing import Optional
-
 import json
 from datetime import datetime, timezone
+from typing import Optional
+
 from botocore.exceptions import ClientError
 from fastapi import APIRouter, HTTPException, Request
 from fastapi.responses import JSONResponse
-from pydantic import BaseModel, Field
-
 from nx_neptune.clients.client_factory import ClientFactory
 from nx_neptune.clients.iam_client import split_s3_arn_to_bucket_and_path
+from pydantic import BaseModel, Field
+
 from nx_neptune_proxy.config import Settings
-from nx_neptune_proxy.routers.schemas import ProjectionExport, ProjectExportPayload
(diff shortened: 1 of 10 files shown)


mykola-pereyma (approved):
LGTM — verified this is formatting-only across all 10 files. Every change is isort import reordering, black line-wrapping, or blank-line normalization. Importantly, the security tests are untouched at the assertion level: in `test_sql_injection.py`, `test_no_secrets.py`, and `test_column_allowlist.py` every payload and assertion is byte-identical — only the surrounding `store.create(...)` calls were wrapped.

One note on `aws_helper.py`, the single non-whitespace change: isort correctly consolidated the module-level duplicate `from fastapi import` (the module previously imported `HTTPException` at the top and `HTTPException, Request` again a few lines down) into one line. Good catch — behavior-preserving.

Small follow-up (optional, and technically beyond "formatting only" so fine to defer): now that `HTTPException` is imported at module scope, the **function-local** `from fastapi import HTTPException` inside `check_key_not_exists` is redundant and could be dropped:

```python
def check_key_not_exists(s3, bucket: str, key: str) -> None:
    """Raise 409 if the S3 key already exists."""
    from botocore.exceptions import ClientError
    from fastapi import HTTPException   # <- now redundant with the module-level import
```

Not blocking — approving as-is since it's a clean, verifiable formatting pass.

PR #433, opened 2026-08-28
Refactor: DB connection context manager + fix test-cases for missing project ID

andy-k-improving:
**Description:**

Refactors DB connection handling into a reusable context manager and fixes the proxy test-cases that were missing a mandatory `project_id`.

- **Optimisation / refactor — `connection()` context manager:** store methods previously called `get_connection()` with manual `close()` and no `try/finally`,
  so a failed query left the connection open holding a SQLite lock and subsequent calls blocked ~5s on it. Added a `connection()` context manager in `db.py` that commits on success and always closes, and converted all call sites to it.

- **Fix test-cases for missing project ID:** tests now supply the mandatory `project_id`.

- **DX:** add a `make proxy-test` target so the proxy suite can be run from repo root
    (mirrors the proxy module's `pytest tests/ -v`).

- **Minor:** `typing.List` annotations to keep mypy / `make proxy-lint` green.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,6 +79,10 @@ proxy-lint:       ## Run flake8, black, mypy linters on the proxy module.
 	$(ENV_PREFIX)black --check proxy/src/ proxy/tests/
 	$(ENV_PREFIX)mypy --ignore-missing-imports proxy/src/
 
+.PHONY: proxy-test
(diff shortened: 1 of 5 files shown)


mykola-pereyma (approved):
LGTM

PR #440, opened 2026-08-28
Feat: Support multiple node and edge queries per projection

andy-k-improving:
## Summary

A projection can now define multiple node queries and multiple edge queries
(previously a single `node_query` / `edge_query`). All are run on import.

## Changes

- **Data model:** new `node_queries` / `edge_queries` tables + `NodeQuery` / `EdgeQuery` models and a `query_store`. `project_id` now required on a projection.

- **API:** `GET`/`POST /api/v0/projection/{id}/queries` to fetch/save the query set; validate, preview, and export/import all handle the full list.

- **Architecture:** new `ProjectionService` layer — routers/pipeline no longer touch the stores directly. DB access centralized on a single `connection()` context manager (fixes a connection leak on failed writes).

- **Bug fix:** deleting a projection/project now also clears its queries (was leaving
  orphaned rows).

- **UI:** node/edge queries are add/remove lists in scrollable frames; editors are
  vertically resizable.

635741508-b1fffd27-1460-49a2-9fe9-e421233d056a

## Testing

- `make proxy-lint` clean, `make proxy-test` **236 passed**, `make ui` builds.

### DIFF
--- a/proxy/src/nx_neptune_proxy/services/db.py
+++ b/proxy/src/nx_neptune_proxy/services/db.py
@@ -38,6 +38,12 @@ def connection() -> Iterator[sqlite3.Connection]:
         conn.close()
 
 
+def query(sql: str, params: tuple = ()) -> list[sqlite3.Row]:
+    """Run a read query and return all rows, always closing the connection."""
+    with connection() as conn:
+        return conn.execute(sql, params).fetchall()
+
+
 def init_db() -> None:
     with connection() as conn:
         conn.executescript("""
--- a/proxy/src/nx_neptune_proxy/services/db.py
+++ b/proxy/src/nx_neptune_proxy/services/db.py
@@ -18,7 +18,12 @@ def get_connection() -> sqlite3.Connection:
     Path(DB_PATH).parent.mkdir(parents=True, exist_ok=True)
     conn = sqlite3.connect(DB_PATH, check_same_thread=False)
     conn.row_factory = sqlite3.Row
-    conn.execute("PRAGMA journal_mode=WAL")
+    # Fetch the result so the pragma statement doesn't stay pending and
+    # swallow the next pragma on some SQLite builds.
+    conn.execute("PRAGMA journal_mode=WAL").fetchone()
+    # Enforce FKs per-connection (off by default), so the ON DELETE CASCADE
+    # in init_db() acts as a real backstop for the explicit query cleanup.
+    conn.execute("PRAGMA foreign_keys = ON")
     return conn
 
 
--- a/proxy/src/nx_neptune_proxy/services/db.py
+++ b/proxy/src/nx_neptune_proxy/services/db.py
@@ -67,4 +67,18 @@ def init_db() -> None:
             created_at TEXT NOT NULL,
             FOREIGN KEY (project_id) REFERENCES projects(id)
         );
+        CREATE TABLE IF NOT EXISTS node_queries (
+            id TEXT PRIMARY KEY,
+            projection_id TEXT NOT NULL,
+            sql TEXT NOT NULL DEFAULT '',
+            position INTEGER NOT NULL DEFAULT 0,
+            FOREIGN KEY (projection_id) REFERENCES projections(id) ON DELETE CASCADE
+        );
+        CREATE TABLE IF NOT EXISTS edge_queries (
+            id TEXT PRIMARY KEY,
+            projection_id TEXT NOT NULL,
+            sql TEXT NOT NULL DEFAULT '',
+            position INTEGER NOT NULL DEFAULT 0,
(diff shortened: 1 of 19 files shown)


proxy/src/nx_neptune_proxy/services/db.py
```diff
@@ -67,4 +67,18 @@ def init_db() -> None:
             created_at TEXT NOT NULL,
             FOREIGN KEY (project_id) REFERENCES projects(id)
         );
+        CREATE TABLE IF NOT EXISTS node_queries (
+            id TEXT PRIMARY KEY,
+            projection_id TEXT NOT NULL,
+            sql TEXT NOT NULL DEFAULT '',
+            position INTEGER NOT NULL DEFAULT 0,
+            FOREIGN KEY (projection_id) REFERENCES projections(id) ON DELETE CASCADE
```
mykola-pereyma, line 80: **[medium] This `ON DELETE CASCADE` (and the FK itself) is inert.** SQLite doesn't enforce foreign keys unless `PRAGMA foreign_keys = ON` is set per connection, and `get_connection()` only sets `PRAGMA journal_mode=WAL` — there's no `foreign_keys` pragma anywhere. So neither the FK constraint nor the cascade does anything:
- The orphan-rows fix actually works only because `ProjectionService.delete()` explicitly clears the queries (`save_node_queries(pid, [])` / `save_edge_queries(pid, [])`). The cascade here is dead code.
- Because the FK isn't enforced on INSERT either, a query row can reference a nonexistent `projection_id` — the declared constraint gives a false sense of integrity.
- Footgun: a future deletion path (or a direct `DELETE FROM projections`) would silently re-introduce orphaned query rows.

Suggest adding `conn.execute("PRAGMA foreign_keys = ON")` in `get_connection()` so the FK + cascade become a real backstop (defense-in-depth alongside the explicit delete). Otherwise, drop `ON DELETE CASCADE` and add a comment that deletion is handled explicitly in the service.

proxy/src/nx_neptune_proxy/services/projection_service.py
```diff
@@ -0,0 +1,149 @@
...
+    def list(self) -> List[Projection]:
+        return projection_store.list()
+
+    def list_by_project(self, project_id: str) -> List[Projection]:
+        return projection_store.list_by_project(project_id)
+
+    def update(self, projection_id: str, **data) -> Optional[Projection]:
+        return projection_store.update(projection_id, **data)
+
+    def delete(self, projection_id: str) -> bool:
```
mykola-pereyma, line 60: **[low] The delete isn't atomic across the query rows and the projection row.** `save_node_queries(pid, [])`, `save_edge_queries(pid, [])`, and the projection-row delete each run in their own `connection()`/transaction. A crash mid-sequence could leave a query-less projection (no orphan, since queries are cleared before the projection row, but a partial state). Same for saving a full set (node-save + edge-save are two transactions). Minor for a local single-user proxy — worth wrapping the multi-step delete/save in a single transaction, or noting it's acceptable.

**[low] Test gap:** I don't see a regression test asserting that deleting a projection clears its queries. Since the `ON DELETE CASCADE` that superficially covers this is inert (see the db.py comment), please add an explicit test so the behavior can't silently regress.

mykola-pereyma (commented):
A couple of non-blocking findings inline (one medium, two low). The feature is solid — SQL is safely parameterized, the `connection()` context manager is a real leak/atomicity fix, and the orphan-rows fix works via explicit deletion. The main thing worth addressing is that the declared `ON DELETE CASCADE` is currently dead code (no `PRAGMA foreign_keys=ON`).

proxy/src/nx_neptune_proxy/services/db.py
andy-k-improving: Added `PRAGMA foreign_keys = ON` in `get_connection()` so the FK and `ON DELETE CASCADE` are now actually enforced as a backstop.

While wiring this up I also found the preceding `PRAGMA journal_mode=WAL` result wasn't being fetched, which was silently swallowing the `foreign_keys` pragma; fetching it fixes that.

proxy/src/nx_neptune_proxy/services/projection_service.py
andy-k-improving: **[low] Non-atomic delete:**
Made `ProjectionService.delete()` run the query deletes and the projection-row delete in a single `connection()`/transaction via new `delete_node_and_edge_queries_on` / `delete_on` helpers, so it can no longer leave a partial state. Kept the explicit child deletion as defense-in-depth alongside the cascade.

**[low] Test gap:**
Added `test_delete_projection_clears_queries` (asserts no orphan rows after an API delete) and `test_fk_cascade_backstop_clears_queries` (deletes the projection row directly to guard the cascade), so the behavior can't silently regress.

mykola-pereyma (approved):
LGTM

PR #441, opened 2026-09-01
Fix: correct stale `IamClient` import in integ tests

andy-k-improving:
### Summary :memo:

Two integ-test files imported `IamClient`, which was renamed to `IamClientWrapper`. This broke test collection with `ImportError`, so all `session_manager` integ tests (and the S3 versioning test) errored before running.

Renamed `IamClient` → `IamClientWrapper` (import + constructor call) in:
- `integ_test/session_manager/conftest.py`
- `integ_test/graph_operations/test_security_s3_versioning.py`

**Tested:** `pytest --collect-only` now collects all 7 tests cleanly (previously failed at conftest import). Constructor signature is unchanged, so it's a drop-in rename.


mykola-pereyma (approved):
LGTM

PR #442, opened 2026-09-02
Feat: Add graph-query endpoint for a projection + query console UI skeleton

andy-k-improving:
### Summary
Backend endpoint to run an **openCypher query** against a projection's graph and return `{ columns, rows }`, plus the **UI skeleton** for the query console (layout only).

### Backend
- `ProjectionService.run_graph_query(id, query, limit)` — resolves the projection, verifies it's queryable (`status == "complete"` + live `graph_id`), runs via `NeptuneAnalyticsClient.execute_query`, returns `{ columns, rows }`.
- `POST /api/v0/projection/{id}/graph-query` — mirrors the existing `/preview` handler.
- **Mutation-keyword denylist** (`CREATE/MERGE/DELETE/DETACH DELETE/SET/REMOVE`, case-insensitive) as a best-effort guardrail applied before execution. **This is not a security control** — it's substring-based and bypassable. Authoritative read-only enforcement requires an IAM-scoped (`ReadDataViaQuery`-only) session, which is **not** in this PR (tracked as follow-up hardening).
- `LIMIT` appended-if-missing + capped, hard row ceiling, query timeout.
- Errors: `ProjectionNotFound`→404, non-queryable/mutation→400, execution failures→sanitized `{ error }` (no ARNs), never 500.

### Frontend (skeleton only)
- `GraphQueryConsole.tsx`: Layout/editor/results shells. Run enabled only when a query exists **and** graph is `available`.

### Testing
- `test_graph_query.py`: 22 unit + endpoint tests. Full backend suite passes; lint clean; UI builds.

### DIFF
--- a/proxy/src/nx_neptune_proxy/routers/projection.py
+++ b/proxy/src/nx_neptune_proxy/routers/projection.py
@@ -21,6 +21,8 @@ from nx_neptune.validators import (
 )
 
 from nx_neptune_proxy.routers.schemas import (
+    GraphQueryRequest,
+    GraphQueryResponse,
     PreviewResponse,
     ProjectionCreate,
     ProjectionResponse,
@@ -33,6 +35,8 @@ from nx_neptune_proxy.routers.schemas import (
 from nx_neptune_proxy.services.pipeline import run_pipeline
 from nx_neptune_proxy.services.projection_service import (
     ProjectionNotFound,
+    ProjectionNotQueryable,
+    ReadOnlyQueryViolation,
     projection_service,
 )
 from nx_neptune_proxy.utils import unpack_query_results
@@ -187,6 +191,35 @@ async def preview_projection(projection_id: str, limit: int = Query(10, ge=1, le
         all_results.append(unpack_query_results(rows))
 
     return {"error": None, "results": all_results}
+
+
+@router.post(
+    "/{projection_id}/graph-query",
+    summary="Run a read-only openCypher query against the projection's graph",
+    response_model=GraphQueryResponse,
+)
+async def graph_query(projection_id: str, body: GraphQueryRequest):
+    """Execute a read-only openCypher query and return ``{columns, rows}``.
+
+    Mirrors the ``/preview`` handler's conventions (same router, same auth,
+    tabular response). A missing projection returns 404; a non-queryable
+    projection or a disallowed mutation returns 400; execution failures are
+    returned as a sanitized ``{error}`` body.
+    """
+    kwargs = {} if body.limit is None else {"limit": body.limit}
+    try:
+        return projection_service.run_graph_query(projection_id, body.query, **kwargs)
+    except ProjectionNotFound:
+        raise HTTPException(status_code=404, detail="Projection not found")
+    except (ProjectionNotQueryable, ReadOnlyQueryViolation) as e:
+        raise HTTPException(status_code=400, detail=str(e))
+    except Exception as e:  # noqa: BLE001 - surface a sanitized message, not a 500
+        return {
+            "error": {
+                "message": sanitize_error_message(str(e)),
+                "hint": "Check the query syntax and that the graph is available.",
+            }
+        }
 
 
 @router.post(
--- a/proxy/src/nx_neptune_proxy/services/projection_service.py
+++ b/proxy/src/nx_neptune_proxy/services/projection_service.py
@@ -16,10 +16,47 @@ tests). Routers translate ``ProjectionNotFound`` into a 404.
 
 from __future__ import annotations
 
-from typing import List, Optional
+import re
+from typing import Any, List, Optional
 
-from .projection_store import Projection, store as projection_store
+from nx_neptune.clients import NeptuneAnalyticsClient
+
+from .projection_store import Projection
+from .projection_store import store as projection_store
 from .query_store import EdgeQuery, NodeQuery, query_store
+
+# --- Read-only graph-query configuration ---
+
+# Projection status at which the backing Neptune graph exists and is queryable.
+QUERYABLE_STATUS = "complete"
+
+# Default LIMIT appended when a query has none, and the hard ceiling on rows
+# returned regardless of any user-supplied LIMIT. These are sensible defaults
+# and are expected to be revisited based on client feedback.
+DEFAULT_LIMIT = 100
+MAX_ROWS = 1000
+
+# Per-query timeout (seconds) applied to the Neptune Analytics client. The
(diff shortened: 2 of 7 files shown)


proxy/src/nx_neptune_proxy/services/projection_service.py
```diff
@@ -30,6 +67,94 @@ def __init__(self, projection_id: str):
...
+    (a read-only-scoped session/role for the Neptune data plane), which cannot
+    be bypassed by query phrasing. This check exists to fail obvious mutations
+    fast with a clear error, not to be a substitute for that IAM scoping.
+
+    ``DETACH DELETE`` is checked first so it is reported in preference to the
+    bare ``DELETE`` it contains.
+    """
+    upper = query.upper()
+    for keyword in _MUTATION_KEYWORDS:
+        if keyword in upper:
```
mykola-pereyma, line 110: **[must fix] Substring match false-positives -- blocks legitimate read queries.** `keyword in upper` is a case-insensitive *substring* test, so ordinary reads are rejected:
- `created_at` contains CREATE, `deleted` contains DELETE, `asset`/`offset`/`subset`/`reset` contain SET, `emergency` contains MERGE, `removed` contains REMOVE.
- e.g. `MATCH (n:Asset) WHERE n.deleted = false RETURN n.created_at` is rejected on three counts.

Use word-boundary matching instead, e.g.:
```python
_MUTATION_RE = re.compile(r"\b(?:DETACH\s+DELETE|CREATE|MERGE|DELETE|SET|REMOVE)\b", re.IGNORECASE)
```
I checked: `\b...\b` no longer matches CREATED_AT / ASSET / DELETED / EMERGENCY / OFFSET / REMOVED. Please also add tests that a read query containing these substrings is ALLOWED -- the current suite only asserts mutations are blocked, so this regression is untested.

proxy/src/nx_neptune_proxy/services/projection_service.py
```diff
@@ -30,6 +67,94 @@ def __init__(self, projection_id: str):
...
+def _ensure_limit(query: str, limit: int) -> str:
+    """Append ``LIMIT <limit>`` when the query has no trailing LIMIT clause.
+
+    A user-supplied LIMIT is left intact; the row ceiling is still enforced on
+    the results after execution as a belt-and-suspenders bound.
+    """
+    stripped = query.rstrip().rstrip(";").rstrip()
+    if _LIMIT_RE.search(query):
+        return query
+    return f"{stripped} LIMIT {limit}"
```
mykola-pereyma, line 124: **[medium] `_ensure_limit` assumes a narrow trailing-`LIMIT <int>` shape and mis-handles others:**
- Parameterized limit `... RETURN n LIMIT $p`: `_LIMIT_RE` requires `\d+`, so it won't match and a second LIMIT is appended -> `LIMIT $p LIMIT 100` (syntax error).
- Trailing line comment `... RETURN n // note`: the `$`-anchored regex won't match, so `LIMIT 100` is appended *into the comment* -> no effective limit (only the post-exec `rows[:MAX_ROWS]` bound applies).

Worth guarding these or documenting the supported query shape. Also: a user-supplied `LIMIT 999999` is left in-query (engine computes it, bounded only by the 30s timeout) -- consider capping the user LIMIT in-query too, not just the appended default.

proxy/src/nx_neptune_proxy/routers/projection.py
```diff
@@ -189,6 +193,35 @@ async def preview_projection(projection_id: str, limit: int = Query(10, ge=1, le
...
+    returned as a sanitized ``{error}`` body.
+    """
+    kwargs = {} if body.limit is None else {"limit": body.limit}
+    try:
+        return projection_service.run_graph_query(projection_id, body.query, **kwargs)
+    except ProjectionNotFound:
+        raise HTTPException(status_code=404, detail="Projection not found")
+    except (ProjectionNotQueryable, ReadOnlyQueryViolation) as e:
+        raise HTTPException(status_code=400, detail=str(e))
+    except Exception as e:  # noqa: BLE001 - surface a sanitized message, not a 500
```
mykola-pereyma, line 216: **[must fix, cheap] The swallowed exception is never logged server-side.** This `except Exception` sanitizes to a client `{error}` (good, no ARNs), but nothing is logged -- and `projection_service.py` has no logger at all. So execution failures are invisible for debugging/ops. Please log the un-sanitized exception here (server-side), ideally with the projection id; and consider basic query duration + row-count metrics for a query-console backend.

Also a contract note: failures are signaled three different ways -- 404, 400, and this 200-with-`{error}`. Clients must check status codes AND inspect a 200 body. Consider making execution failures a 4xx too (or documenting why they're 200).

proxy/ui-src/src/components/GraphQueryConsole.tsx
```diff
@@ -0,0 +1,180 @@
...
+  meta,
+  onClose,
+}: {
+  meta: GraphMeta;
+  onClose: () => void;
+}) {
+  const [query, setQuery] = useState(DEFAULT_QUERY);
+  // Result/error state and the render shells below are kept so wiring the API
+  // response only needs to populate them.
+  const [result] = useState<ResultTable | null>(null);
```
mykola-pereyma, line 40: **[medium] Dead state / non-functional Run shipped into the real page.** `result` and `error` are declared with `useState` but no setter (permanently `null`), and `handleRun` is a TODO -- so Run does nothing. That's fine as a skeleton, but `GraphQueryConsole` is actually rendered in `Projections.tsx` behind a "Query Graph" button, so end users can open the console, type a query, click Run, and get no response. Please gate it behind a feature flag (or don't render it) until it's wired. Note once wired, the denylist false-positives above would surface as confusing "disallowed keyword: SET" errors on normal queries.

mykola-pereyma (changes requested):
The backend is cleanly decomposed and the security posture is honestly disclosed. Requesting changes on two must-fix items and flagging a few design/UX points inline.

Must-fix: (1) the mutation denylist's substring matching rejects ordinary read queries (created_at, deleted, asset...) -- switch to word boundaries + add allow-tests; (2) the swallowed execution exception is never logged server-side.
Discuss: the tri-modal error contract (404/400/200-with-error), `_ensure_limit` edge cases, and shipping a non-functional Run button into the Projections page.

Security posture (denylist is best-effort/bypassable; IAM read-only scoping deferred) is acceptable on this P2 branch given the token+localhost gating and the honest disclosure -- but the IAM ReadDataViaQuery scoping should be a hard gate before GA, since this passes arbitrary openCypher to a write-capable session.

PR #448, opened 2026-09-04
Enhance: escape node/edge property keys in openCypher query builder

andy-k-improving:
**Summary**
  
  Extends the query builder's identifier-escaping to cover property keys (attribute names), completing the escaping already applied to labels, relationship types, and algorithm-parameter keys. Property values and IDs continue to be parameterized ($N); this rounds out consistent handling of the identifier side of the query.
  
**What changed**
  
  Previously ParameterMapBuilder.read_map masked values as $N but emitted keys as-is. Since openCypher has no parameter placeholder for identifier positions, keys are now backtick-escaped in the builder the same way labels already are — so any attribute name (including unusual ones with special characters) is emitted as a well-formed identifier.
  
  nx_neptune/clients/opencypher_builder.py:
  
  - Escape keys at the single read_map choke point.
  - New KeyStyle enum for the two key shapes the builder produces:
    - PLAIN (default) — bare property names (CREATE/MERGE maps): escaped wholesale.
    - PATH — SET/WHERE keys: a.age → a.`age` (escape only the property segment); id(n) predicates pass through unchanged; anything unexpected is escaped defensively.
  
  - Removed now-redundant `~id` pre-backticking (the choke point handles it).
  - Applied PATH at the reference/predicate callers (update_node/edge, bfs_query, descendants_*, bfs_layers_query, jaccard_*).
  
**Testing**
  
  - New TestPropertyKeyInjection + escaper-helper tests (test_opencypher_injection.py).
  - New integration test test_security_property_key_injection.py — round-trips unusual/edge-case attribute names against a real graph, with positive controls confirming plain keys, dotted
  ref.prop references, and id(n) predicates still behave correctly.
  - Updated expected-query fixtures (test_opencypher_builder.py, test_parameter_map_builder.py).
  - Full unit suite 443 passed; flake8 + black clean.

### DIFF
--- a/integ_test/graph_operations/test_security_property_key_injection.py
+++ b/integ_test/graph_operations/test_security_property_key_injection.py
@@ -1,0 +1,211 @@
+# Copyright 2025 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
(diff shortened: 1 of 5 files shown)


mykola-pereyma (approved):
LGTM

PR #457, opened 2026-09-10
Misc: Enhance table_columns validation logic

andy-k-improving:
## Changes

**`nx_neptune/instance_management.py`**
- `create_iceberg_table_from_table`: validate each `table_columns` entry with `_validate_sql_identifier` before it's interpolated into the `AS SELECT` list.
- `create_csv_table_from_s3`: removed the unused `table_columns` param so it can't become a second unvalidated sink.


mykola-pereyma (approved):
LGTM

PR #466, opened 2026-09-16
Enhancement: scope notebook Glue permissions and disable unnecessary access

andy-k-improving:
## Summary
Tightens the SageMaker notebook role's Glue Data Catalog permissions to least-privilege and disables root access on the notebook instance for the demo stack. 

## Changes
- Split the combined `GlueDataCatalog` statement into `GlueReadCatalog` (reads) and `GlueCreateTable` (create)
- Removed `glue:UpdateTable`, `glue:DeleteTable`, `glue:CreateDatabase` (unused by the import/export flow)
- Scoped `glue:CreateTable` to `database/${ApplicationId}*` and `table/${ApplicationId}*/*` (used by the Iceberg export path)

## Testing
- CloudFormation template validated as well-formed JSON
- Confirmed the export path (`create_iceberg_table_from_table`) is the only Glue writer and needs only `CreateTable`; data read/write is handled by the separate S3 + KMS statements


cloudformation-templates/nx-neptune-sagemaker.json
```diff
@@ -300,24 +300,31 @@
...
-              "Sid": "GlueDataCatalog",
+              "Sid": "GlueReadCatalog",
               "Effect": "Allow",
               "Action": [
                 "glue:GetDatabase",
+                "glue:GetDatabases",
                 "glue:GetTable",
                 "glue:GetTables",
-                "glue:CreateTable",
-                "glue:UpdateTable",
```
acarbonetto, line 333: Please confirm the following:  
1. glue:CreateTable mis-scoped (nx-neptune-sagemaker.json:322) — Scoped to database/${ApplicationId}* and table/${ApplicationId}*/*, but the code creates tables in default (OnStart sets NETWORKX_S3_TABLES_DATABASE='default', line 365) and the example uses bank_fraud_full. No Glue DB is ever named after ApplicationId, so CreateTable is denied → export flow fails with AccessDenied.
2. glue:UpdateTable removal breaks Iceberg export (:305) — create_iceberg_table_from_table runs CREATE TABLE ... table_type='ICEBERG' AS SELECT. Athena Iceberg commits data by updating the Glue table's metadata_location via glue:UpdateTable. Removing it lets the CTAS create succeed but the data-commit step fails. Contradicts the PR's claim that UpdateTable is unused.

andy-k-improving: **1. `CreateTable` mis-scoped — confirmed.** Export writes to `default` (from `NETWORKX_S3_TABLES_DATABASE`), which never matched `${ApplicationId}*`. Added a `GlueDatabaseScope` param (default `default`), scoped the write ARNs to it, and wired the notebook's `NETWORKX_S3_TABLES_DATABASE` to the same param so they can't drift.

**2. `UpdateTable` removal — confirmed.** Iceberg CTAS commits via the table's `metadata_location`, which needs `glue:UpdateTable`. Restored it; the statement (now `GlueWriteTable`) grants `CreateTable` + `UpdateTable`, scoped to `GlueDatabaseScope`. `DeleteTable`/`CreateDatabase` stay removed. README updated.

acarbonetto (approved):
Thanks for verifying those questions.  LGTM.

PR #467, opened 2026-09-16
Enhancement: secure-by-default graph connectivity and deletion protection

andy-k-improving:
## Summary
Makes the SageMaker demo stack secure by default — the graph is private and deletion-protected unless explicitly relaxed for a convenient demo.

## Changes
- `PublicConnectivity` parameter default flipped `true` → `false` (private, VPC-only)
- Added a `DeletionProtection` parameter (default `true`), replacing the hardcoded `false` on the graph via an `EnableDeletionProtection` condition
- `deploy.sh`: added `PUBLIC_CONNECTIVITY` / `DELETION_PROTECTION` env overrides (secure defaults; e.g. `PUBLIC_CONNECTIVITY=true DELETION_PROTECTION=false ./deploy.sh` for the demo)
- `teardown.sh`: disables deletion protection on the graph (via `neptune-graph update-graph` on the `GraphId` output) before `delete-stack`, so teardown still succeeds


cloudformation-templates/nx-neptune-sagemaker.json
```diff
@@ -16,7 +16,13 @@
       "AllowedValues": ["16", "32", "64"]
     },
     "PublicConnectivity": {
-      "Description": "Enable public connectivity for the graph",
+      "Description": "Enable public connectivity for the graph (default false = private, VPC-only)",
+      "Type": "String",
+      "Default": "false",
```
mykola-pereyma, line 21: Flipping this to `false` by default is reasonable, but it makes the default deploy non-functional as the template stands: the `NotebookInstance` isn't in a VPC and there's no `PrivateGraphEndpoint`. A Neptune Analytics graph with public connectivity off is only reachable from inside its VPC, and both notebook paths need that per-graph host — the `%%oc` magic (`GRAPH_NOTEBOOK_HOST=Graph.Endpoint`, 8182) and the `boto3` `execute_query` call (a data-plane op served from the same per-graph host, not a regional API). So on the defaults the notebook can't reach the graph and users would set `PUBLIC_CONNECTIVITY=true` to recover.

Please pair the private default with the notebook placed in the graph's VPC/subnet(s) + a `PrivateGraphEndpoint` (SG ingress on 443, and 8182 for `%%oc`) so the default path works end-to-end.

mykola-pereyma (changes requested):
The `DeletionProtection` changes look solid — the param + `Fn::If` on the graph + the `teardown.sh` `update-graph --no-deletion-protection` step all hang together, and I confirmed the template exports the `GraphId` output that teardown resolves. No concerns there.

Requesting changes on the `PublicConnectivity` default flip, though — it's the right instinct, but as written the **default deploy won't be able to reach the graph**. The bundled `NotebookInstance` isn't placed in a VPC, and there's no `AWS::NeptuneGraph::PrivateGraphEndpoint` in the template. With `PublicConnectivity=false`, a Neptune Analytics graph's endpoint is only reachable from inside the graph's VPC, and both of the notebook's connection paths depend on that host:
- the graph-notebook `%%oc` magic points `GRAPH_NOTEBOOK_HOST` at `Graph.Endpoint` (port 8182), and
- the `nx_neptune` `boto3` `execute_query(graphIdentifier=...)` call is a data-plane op served from the *same* per-graph host (`g-<id>.<region>.neptune-graph.amazonaws.com`), not a regional endpoint.

So on the default settings the notebook can't connect, and anyone running the demo would have to set `PUBLIC_CONNECTIVITY=true` to get a working graph — which lands the default right back where it started.

To make private-by-default actually usable, the connectivity plumbing needs to come with it: place the `NotebookInstance` in the graph's VPC/subnet(s) and add a `PrivateGraphEndpoint` (with security-group ingress on 443, plus 8182 if you want the `%%oc` magic to work). With that in place the secure default is functional end-to-end. Happy to re-review once that's wired up.

mykola-pereyma (approved):
Updating my earlier review to approve — on a closer read the secure defaults are handled well end-to-end: the CloudFormation template now defaults PublicConnectivity=false and DeletionProtection=true (via the param + Fn::If), the library create paths already default the same way (private + protected, with env opt-in/opt-out), the teardown script clears deletion protection before delete-stack, and the README's "Connecting to a private graph (VPC)" section documents the private-endpoint path. Verified the template exports the GraphId output that teardown relies on. LGTM.

Two non-blocking nits:

1. `cloudformation-templates/README.md` (the parameter table, ~line 87) still lists `PublicConnectivity` default as `true`. This PR flips the actual template default to `false`, so that doc line is now stale — please update it to `false` to match.

2. (optional) With `PublicConnectivity=false` the graph is reachable only from inside its VPC, but the SageMaker template doesn't place the NotebookInstance in a VPC or provision a `PrivateGraphEndpoint` — so on the secure defaults the bundled notebook can't reach the graph and a demo user would set `PUBLIC_CONNECTIVITY=true` to recover. The README documents the VPC path as the operator's responsibility, so this isn't blocking; but wiring the notebook into the graph's VPC/subnet + a `PrivateGraphEndpoint` (SG ingress 443, plus 8182 for the `%%oc` magic) would make the secure default work out of the box.

PR #468, opened 2026-09-16
Enhancement: scope NotebookRole trust policy to the deploying account

andy-k-improving:
## Summary
Adds cross-service confused-deputy conditions to the `NotebookRole` trust policy so the SageMaker and Neptune Analytics service principals can only assume the role on behalf of resources in the stack's own account and region.

## Changes
- Added a `Condition` to the `sagemaker.amazonaws.com` statement: `aws:SourceAccount = ${AWS::AccountId}` and `aws:SourceArn ArnLike arn:${AWS::Partition}:sagemaker:${AWS::Region}:${AWS::AccountId}:notebook-instance/*`

- Added a `Condition` to the `neptune-graph.amazonaws.com` statement: `aws:SourceAccount = ${AWS::AccountId}` and `aws:SourceArn ArnLike arn:${AWS::Partition}:neptune-graph:${AWS::Region}:${AWS::AccountId}:graph/*`


cloudformation-templates/nx-neptune-sagemaker.json
```diff
@@ -135,12 +135,28 @@
             {
               "Effect": "Allow",
               "Principal": { "Service": "sagemaker.amazonaws.com" },
-              "Action": "sts:AssumeRole"
+              "Action": "sts:AssumeRole",
+              "Condition": {
+                "StringEquals": { "aws:SourceAccount": { "Ref": "AWS::AccountId" } },
+                "ArnLike": {
+                  "aws:SourceArn": {
+                    "Fn::Sub": "arn:${AWS::Partition}:sagemaker:${AWS::Region}:${AWS::AccountId}:notebook-instance/*"
```
mykola-pereyma, line 143: This `notebook-instance/*` source-ARN is the main thing to verify against a live deploy. AWS documents no confused-deputy example for notebook instances, and for the execution-role cases it does document it uses an account-scoped `arn:aws:sagemaker:<region>:<acct>:*` (the `*` is literal, not a wildcard). If SageMaker sends that shape when assuming the notebook execution role, this `ArnLike ...:notebook-instance/*` won't match — and since a positive `ArnLike` on a missing/mismatched key evaluates false, the statement flips to an implicit deny and the notebook won't launch. Please confirm the notebook still starts with this in place; if not, use the documented `...:<acct>:*` form or an `...IfExists` operator.

mykola-pereyma (changes requested):
Thanks — scoping the trust policy to the deploying account/region is the right direction, and the two statements are wired correctly (`aws:SourceAccount` + an `ArnLike aws:SourceArn`, partition/region/account via `Fn::Sub`).

Requesting changes for one reason: these conditions need to be confirmed against a real deploy before merge, because a cross-service condition only works if the calling service actually populates `aws:SourceArn`/`aws:SourceAccount` on the `sts:AssumeRole` call — and per IAM's rules a positive `StringEquals`/`ArnLike` on a **missing** key evaluates false, which flips the statement to an implicit **deny**. So if either principal doesn't send the source keys in the shape used here, the role can no longer be assumed and the stack breaks.

Two specific worries:
- **SageMaker (`notebook-instance/*`)**: AWS publishes confused-deputy examples for Inference/Studio/Training/etc. but none for notebook instances, and for the execution-role cases it does document it uses an *account-scoped* source ARN (`arn:aws:sagemaker:<region>:<acct>:*`, where the trailing `*` is literal, not a wildcard). If SageMaker sends that shape for a notebook execution role rather than `...:notebook-instance/<name>`, the `notebook-instance/*` `ArnLike` won't match and the notebook won't launch.
- **Neptune (`graph/*`)**: AWS's documented import/export assumed-role trust policy carries no condition block, and it isn't documented that the service sends `aws:SourceArn=graph/<id>` for StartImportTask/StartExportTask/CreateGraphUsingImportTask. Same missing-key deny risk if it doesn't.

Could you confirm a live deploy with these in place — specifically that (1) the SageMaker notebook instance launches, and (2) a StartImport/StartExport task succeeds? The PR description doesn't mention any testing, and this is the kind of change that fails closed. If either is denied, the safe fallbacks are the AWS-documented account-scoped source ARN (`...:<acct>:*` exactly) or `...IfExists` operators so a missing key doesn't deny. Happy to re-review once it's deploy-verified.

cloudformation-templates/nx-neptune-sagemaker.json
acarbonetto: Please validate with notebooks before merging.

cloudformation-templates/nx-neptune-sagemaker.json
```diff
@@ -135,12 +135,28 @@
...
             {
               "Effect": "Allow",
               "Principal": { "Service": "neptune-graph.amazonaws.com" },
-              "Action": "sts:AssumeRole"
+              "Action": "sts:AssumeRole",
+              "Condition": {
+                "StringEquals": { "aws:SourceAccount": { "Ref": "AWS::AccountId" } },
+                "ArnLike": {
+                  "aws:SourceArn": {
+                    "Fn::Sub": "arn:${AWS::Partition}:neptune-graph:${AWS::Region}:${AWS::AccountId}:graph/*"
```
acarbonetto: wildcarding the resource may be safer.  We should consider wildcarding the resource type to `arn:${AWS::Partition}:neptune-graph:${AWS::Region}:${AWS::AccountId}:*` which should be equally safe.  

Please validate with a live import/export run to ensure that the tasks aren't going to return an access denied.

cloudformation-templates/nx-neptune-sagemaker.json
andy-k-improving: Deploy-verified on `nx-nb-p1test` (us-west-1) — both work, neither fails closed.
SageMaker assumed the role (notebook reached `InService`), and Neptune assumed it too. 
Left the strict form in place.

cloudformation-templates/nx-neptune-sagemaker.json
andy-k-improving: Done and validated. Broadened the Neptune trust-policy `aws:SourceArn` to `arn:${AWS::Partition}:neptune-graph:${AWS::Region}:${AWS::AccountId}:*` as suggested — equally safe (the account+region pin enforces the confused-deputy protection; the resource-type segment adds no real defense against a cross-account caller) and more robust to the `import-task/*`/`export-task/*` shapes the service may present.

Permission policy left untouched (`graph/*`, `import-task/*`, `export-task/*` unchanged). Tested with live `StartImportTask` on the deployed formation stack.

acarbonetto (approved):

PR #481, opened 2026-09-19
Add CREATE PROPERTY GRAPH front-end for Athena projections

cornerwings:
## Summary

Adds a schema-first front-end for projecting data-lake tables into Neptune Analytics. Instead of hand-writing the Athena `SELECT` statements that alias source columns into the Gremlin CSV load format (`~id`/`~label` for vertices, `~from`/`~to`/`~label` for edges), callers can declare the mapping with the standard SQL/PGQ (SQL:2023) `CREATE PROPERTY GRAPH` DDL. The DDL is translated into those same projection queries and runs through the unchanged Athena → S3 → Neptune import path — only the front-end changes.

## Motivation

The existing `import_from_table` flow requires users to author the projection SQL by hand for every source, which is repetitive and easy to get subtly wrong (label literals, `:Type` header suffixes, null filters). Declaring a property-graph schema is more readable, less error-prone, and aligns the library with a standard schema language.

## What's included

- `nx_neptune/property_graph.py` — a lark grammar + transformer that parses `CREATE PROPERTY GRAPH` DDL into a small schema model, plus a SQL generator (`property_graph_to_sql`, `parse_property_graph`).
- `SessionManager.import_from_graph_schema(...)` — new method that translates the DDL and delegates to the existing `import_from_table`.
- Public exports from `nx_neptune`.
- `lark` added to `pyproject.toml` and regenerated lock files (`make lock`).
- Docs: README section, a new `Property Graph Schema` docs page (+ sidebar entry), and a section on the Session Manager page.
- A runnable example: `example/instance_management/property_graph_schema_import.py`.
- Unit tests in `tests/test_property_graph.py`.

## Example

```python
DDL = """
CREATE PROPERTY GRAPH financial
  VERTEX TABLES (
    accounts AS customer KEY (name) LABEL customer PROPERTIES (name)
  )
  EDGE TABLES (
    transactions
      SOURCE KEY (nameOrig) REFERENCES customer
      DESTINATION KEY (nameDest) REFERENCES customer
      LABEL transfer
      PROPERTIES (step:Int, amount:Float, isFraud:Int)
  )
"""

await session.import_from_graph_schema(graph, s3_location, DDL, catalog=catalog, database=database)
```

The optional `:Type` suffix (e.g. `amount:Float`) matches the load-format header convention; `LABEL` defaults to the table name/alias when omitted. `property_graph_to_sql(DDL)` returns the generated SQL without importing.

## Limitations (this PR)

- Keys must be single-column (a Neptune `~id`/`~from`/`~to` maps to one column); composite keys are rejected with a clear error.
- `PROPERTIES ARE ALL COLUMNS` is not supported (would require Athena schema introspection to emit typed headers).

## Testing

- `tests/test_property_graph.py` — 16 tests covering translation, label defaulting, property rename/type suffixes, `NO PROPERTIES`, comments, dotted/quoted identifiers, quoting/escaping, and all rejection paths.
- Full suite: 412 passed locally.


acarbonetto:
Thank you so much @cornerwings.  Would you mind fixing the CI/linter while I take a look at the changes.

cornerwings:
Thanks @acarbonetto, fixed up the linter issues - will address feedback once you take a look at changes.

